PR #1, opened 2016-09-01
File (MasterNode) and class (Masternode) name unification

kilerixs:
File (MasterNode.java) and class (Masternode) name unification


HashEngineering:
This issue should have been resolved over a year ago.

PR #3, opened 2016-09-01
Helper method to check transaction is dust

kilerixs:
Helper method to check transaction is dust


HashEngineering:
What is the reasoning behind this change?  Was it established by Dash Core?

HashEngineering:
This method is already in our code.

PR #4, opened 2016-11-22
Fixed a bug causing NPE in countEnabled(int) method.

tomasz-ludek:
Hi @HashEngineering,
I came across this nasty bug last week. In my case it caused a NullPointerException at [line 429 of MasternodeManager class](https://github.com/HashEngineering/dashj/blob/master/core/src/main/java/org/bitcoinj/core/MasternodeManager.java#L429), but the **context** field is used also in few other places, so it can be even more dangerous.


HashEngineering:
I will check to see if this is still an issue in our code.

HashEngineering:
We pass the Context object as a parameter in the MasternodeManager constructor, so this is no longer needed.

PR #16, opened 2018-05-10
Masternode Messages for Dash 12.2

HashEngineering:
First part is focusing on 
mnb
mnp

to load the masternode list before other messages are added.


HashEngineering:
This was replaced with the develop branch.

PR #19, opened 2018-08-09
Transaction States

HashEngineering:
Modified TransactionConfidence to include:

Time
Number of connected peers
Minimum required connected peers

When a transaction is sent.  This information will be displayed using TransactionConfidence.toString.

A SENT TransactionConfidence.ChangeReason was added to signal that a transaction has been sent with Peer.sendMessage.  An app can use this information to signal the user that the transaction was sent.  Ideally the app should still wait for connected peers to announce the tx with "inv".  However in the case that only one peer is connected, then this SENT reason can be used to indicate some success to the user.

Additionally, after a transaction is locked (when using InstantSend), then the wallet is saved.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>org.dashj</groupId>
         <artifactId>dashj-parent</artifactId>
-        <version>0.14.4.5</version>
+        <version>0.14.4.6</version>
     </parent>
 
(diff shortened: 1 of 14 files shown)


HashEngineering:
This pull request does not include a new version commit, but will after testing and review is complete.

HashEngineering:
This has passed QA.

PR #24, opened 2019-02-01
Fix bug found when downloading full blocks

HashEngineering:


HashEngineering:
This was already merged.

PR #25, opened 2019-02-01
update bip32Header values for dash

pvyhnal-generalbytes:


pvyhnal-generalbytes:
Cc @GENERALBYTES, @rpanak-generalbytes

pvyhnal-generalbytes:
ok, closing this. We are using dpub & dprv but we can keep it in our fork

Issue #29, opened 2019-02-27
Deserialization failure


jdojff:
Hi did not check deep into it what is wrong, but on livenet some of the dashj (0.15) instances are unable to sync due to the similar errors. Any info regarding this case?

```
2019-02-27 22:10:38.730 WARN  [NioClientManager] PeerSocketHandler:234 - [111.111.111.111]:9999 - : org.bitcoinj.core.ProtocolException: Error deserializing message 03000600000000000000fd49010100aab00f000100019cfc7ec999e13a421207b26e0d3e0561db4fed681c99773c3100000000000000320000000000000032000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000

        at org.bitcoinj.core.BitcoinSerializer.deserializePayload(BitcoinSerializer.java:210) [dashj-core-0.15.jar:]
        at org.bitcoinj.core.BitcoinSerializer.deserialize(BitcoinSerializer.java:171) [dashj-core-0.15.jar:]
        at org.bitcoinj.core.PeerSocketHandler.receiveBytes(PeerSocketHandler.java:155) [dashj-core-0.15.jar:]
        at org.bitcoinj.net.ConnectionHandler.handleKey(ConnectionHandler.java:223) [dashj-core-0.15.jar:]
        at org.bitcoinj.net.NioClientManager.handleKey(NioClientManager.java:86) [dashj-core-0.15.jar:]
        at org.bitcoinj.net.NioClientManager.run(NioClientManager.java:122) [dashj-core-0.15.jar:]
        at com.google.common.util.concurrent.AbstractExecutionThreadService$1$2.run(AbstractExecutionThreadService.java:60) [guava-18.0.jar:]
        at com.google.common.util.concurrent.Callables$3.run(Callables.java:95) [guava-18.0.jar:]
        at org.bitcoinj.utils.ContextPropagatingThreadFactory$1.run(ContextPropagatingThreadFactory.java:49) [dashj-core-0.15.jar:]
        at java.base/java.lang.Thread.run(Thread.java:834) [java.base:]
Caused by: java.lang.NullPointerException
[...]

Issue #28, opened 2019-02-27
possible memory leaks when not initialized properly


pvyhnal-generalbytes:
When dashj is initialized with just `Context.get().initDash(false, true);` it leaks memory as `MasternodeManager#mapSeenMasternodePing` is populated but never cleaned.

`Context.get().setLiteMode(false);` has to be called to start ThreadCheckDarkSendPool which cleans the map.

Also `mapTxLockVotes` seems to be getting filled with entries that will never be removed because `TransactionLockVote#isExpired` is never true on them

This seems to be a problem only when lite mode = false. I tried using the lite mode but it had some problems discovering peers. What exactly is is the non-lite mode used for?

Thanks

pvyhnal-generalbytes:
it seems that even if ThreadCheckDarkSendPool is started the way I described above it stops after some time and the memory leak is back. It's probably better to use the lite mode.

HashEngineering:
LiteMode is to run a node without trying to get the list of masternodes, governance objects and governance votes. When LiteMode is OFF then the node tries to get a list of all of these other Dash objects.  OFF is experimental and will soon become partially obsolete with the activation of Spork 15.   

The Dash Wallet android app uses LiteMode and does not have issues finding peers.  There is a fix for finding peers that will be added.  I will comment later with the commit of that fix.

HashEngineering:
Based on your first comment, it would seem that you want litemode ON as it seems to solve most of the problems.  Additionally, in our next release of dashj 0.15.1, we will disable acquisition of masternode objects.  This will resolve the memory leak of this:  `MasternodeManager#mapSeenMasternodePing`, but won't address other issues.  The other issues will need to be looked at further to determine why `mapTxLockVotes` is never emptied while litemode is OFF. 

https://github.com/dashevo/dashj/pull/30/commits/6ffd45ccb02ae3e08151a3aabf721e0eef344510 will address connectivity issues.  Older nodes such as 0.12.3.x will not connect to dashj.

https://github.com/dashevo/dashj/pull/30/commits/7eda2cea72b48ba25177bb9e6338eb3e18f0c257 will ensure that dashj follows the correct chain (the same as the 0.13.x nodes).

HashEngineering:
We are investigating the memory leaks with `mapTxLockVotes` and found that in full verification mode, the `confirmedHeight` is not initialized, therefore `TransactionLockVote.isExpired` never returns true.  This applies to several other `InstantSend` objects.

[...]

PR #30, opened 2019-03-04
DIP3 Activation Updates

HashEngineering:
This will not allow connections to older nodes and will ensure that the correct chain is followed.

This also resolves memory leaks with MasternodeManager (by disabling obsolete message processing, by default) and other leaks that pertained to InstantSend.

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/MasternodeSync.java
+++ b/core/src/main/java/org/bitcoinj/core/MasternodeSync.java
@@ -108,10 +108,19 @@ public class MasternodeSync {
             this.syncFlags = EnumSet.of(SYNC_DMN_LIST);
         else {
             this.syncFlags = EnumSet.noneOf(SYNC_FLAGS.class);
-            syncFlags.add(SYNC_MASTERNODE_LIST);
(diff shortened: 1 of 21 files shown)


HashEngineering:
This addresses:
#29 
#28

core/src/main/java/org/bitcoinj/core/MasternodeSync.java
```diff
@@ -108,13 +108,22 @@ public MasternodeSync(Context context)
             this.syncFlags = EnumSet.of(SYNC_DMN_LIST);
         else {
             this.syncFlags = EnumSet.noneOf(SYNC_FLAGS.class);
-            syncFlags.add(SYNC_MASTERNODE_LIST);
-            syncFlags.add(SYNC_GOVERNANCE);
-            syncFlags.add(SYNC_DMN_LIST);
+            //TODO:add other flags here to get other information such as governance messsages, by default
```
HashEngineering: #28 This should eliminate memory leaks with respect to MasternodeManager, which will not be used if one of the above flags is not set.

PR #36, opened 2019-04-04
BU derivation path

HashEngineering:
Moved from the other dashj branch.

The derivation path used by iOS has changed and another change is needed to get the path to work the same as iOS.  That will have to be added later.


HashEngineering:
This PR is now obsolete.  It has been superseded by #37.

HashEngineering:
This is now obsolete and will be closed.

PR #37, opened 2019-04-09
Add authentication keys to the wallet and include them in the bloom filters

HashEngineering:


HashEngineering:
This is already a part of #40, though there is one more commit to be added.

PR #41, opened 2019-05-12
Authentication Key Chains - Serialization / Deserialization

HashEngineering:
This does not support BLS Key Chains.

This does not keep track of how many keys have been issued.  At first we only need one.  If the user wants to "reset" the blockchain user key, then this will not keep track of that.

Additionally, we don't have a standard scheme for that.  Would we use '1'th key instead of '0'?


HashEngineering:
This needs more work based on what we are doing in #49

HashEngineering:
This may be replaced by #41 as much was learned.

PR #44, opened 2019-06-24
State transition constructor

sambarboza:


HashEngineering:
This would get merged in 0.16.1

core/src/main/java/org/bitcoinj/evolution/SubTxTransition.java
```diff
@@ -42,6 +58,35 @@ protected void parse() throws ProtocolException {
         length = cursor - offset;
     }
 
+    @Override
```
HashEngineering: I don't see this method as being necessary and it doesn't follow the conventions of Message.unsafeBitcoinSerialize because it serializes data that is not included in this object.

Can you instead use Transaction.unsafeBitcoinSerialize with a Transaction object that has the extra payload data of type SubTxTransition.

```
Transaction tx = new Transaction(params);
tx.setExtraPayload(new SubTxTransition(...));
byte [] bytes = tx.unsafeBitcoinSerialize();
```

What is the reason for having this function?  What are you doing with the data?

core/src/main/java/org/bitcoinj/wallet/SendRequest.java
```diff
@@ -237,6 +238,19 @@ public static SendRequest forSubTxResetKey(NetworkParameters params, SubTxResetK
...
+    /**
+     * <p>Creates a new SendRequest for a SubTxTransition (State Transition) with credits.</p>
+     * The SubTxTransition payload must already be signed.
+     */
+    public static SendRequest forSubTxTransition(NetworkParameters params, SubTxTransition subTxTransition) {
+        SendRequest req = new SendRequest();
+        req.tx = new Transaction(params);
+        req.tx.setVersion(3);
+        req.tx.setType(Transaction.Type.TRANSACTION_SUBTX_TRANSITION);
+        req.tx.setExtraPayload(subTxTransition);
```
HashEngineering: These two lines setting the version and type are redundant because that was added to `Transaction.setExtraPayload`

HashEngineering:
This PR is now obsolete given that state transitions will not be happening on the main chain.

PR #56, opened 2019-09-30
LLMQBackgroundThread:  Reduce memory usage

HashEngineering:
InstantSendManager and SigningManager both have process pending methods
that were allocating memory when there was nothing to process.


HashEngineering:
This can also go in the release-0.16 branch

HashEngineering:
This passed QA on the 0.16.4 release.

PR #57, opened 2019-10-02
SimplifiedMasternodeListManager: Handle other exceptions when loading bootstrap

HashEngineering:
Some exceptions were found to occur on some android versions that
resulted in crashes.  This will catch those.

This would also need to be `cherry-pick`ed into the release-0.16 branch if needed for 0.16.4


HashEngineering:
This passed QA on the 0.16.4 release.

PR #58, opened 2019-10-04
Reduce max cache size

denlb:
This fix should help with Krip phones

### DIFF
--- a/core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
+++ b/core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
@@ -34,7 +34,7 @@ public class SimplifiedMasternodeListManager extends AbstractManager {
     public static final int LISTS_CACHE_SIZE = 576;
     public static final int SNAPSHOT_TIME_PERIOD = 60 * 60 * 26;
 
-    public static final int MAX_CACHE_SIZE = 10;
+    public int MAX_CACHE_SIZE = 8;
 
     public enum SaveOptions {
         SAVE_EVERY_BLOCK,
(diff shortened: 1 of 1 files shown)


HashEngineering:
Great, I will test this.  If it works, we can think about using a global constant/setting in `Context` or some other means that can be controlled by the app/program that uses dashj.

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
```diff
@@ -91,6 +91,9 @@ protected boolean removeEldestEntry(Map.Entry<Sha256Hash, SimplifiedQuorumList>
 
     public SimplifiedMasternodeListManager(Context context) {
         super(context);
+        if (Runtime.getRuntime().totalMemory() <= Math.pow(1024, 3)) {
+            MAX_CACHE_SIZE = 1;
+        }
```
HashEngineering: Runtime.getRuntime().totalMemory() does not appear to get the total amount of device memory. 
It returned 256 MB on a 16 GB machine and 8MB on a 1 GB Krip phone.

besides that, MAX_CACHE = 1 appears to work as far as instantsend/chainlock functionality.  I think we should make MAX_CACHE = 1 as static final since keeping 8 or 10 lists in memory on other higher memory devices is unnecessary.  Therefore there is no need to check for device memory size.  A separate task will be created to perhaps keep 2 lists in memory (the 8th previous and the 7th previous) or some other optimization for higher memory devices.  This would address issues where a chainlock or instantsendlock arrives before the DMN List is updated.

HashEngineering: I will do this change.

PR #60, opened 2019-10-19
Add BLS Integrated Encryption Scheme Support

HashEngineering:
Other changes may be required depending on if this lines up with the DashSync-iOS implementation.


HashEngineering:
@QuantumExplorer reviewed the test here https://github.com/dashevo/dashj/pull/60/files#diff-8f0b838a61c218790746694e5d8f23adR77 and found that it passes for the iOS implementation.

HashEngineering:
This can be squashed and merged.

PR #62, opened 2019-11-08
Use BLS Lazy Public keys on DMN Lists

HashEngineering:
The public keys are only used to verify quorums every 24 blocks.

A public key uses 152 bytes of native memory = about 0.7 MB for 5000 masternodes.

This memory is largely unused.  The idea behind lazy public keys is to store a public key only as the 48 bytes until it needs to be used to verify a quorum.  After that this implementation keeps the public key in memory as a BLSPublicKey (using java and native memory) and allows the 48 bytes to be freed up.

Perhaps a further optimization for reducing memory would be to always convert the 48 bytes to public keys each time the key is required, which is not often in SPV mode.


HashEngineering:
Good call on the unit test.

PR #67, opened 2020-02-24
Remove DIP5 special transactions [NMA-389]

HashEngineering:
Remove DIP5 special transactions for Blockchain Users

Remove logic of Blockchain User Transactions


HashEngineering:
This PR will remove DIP5 special transaction support for Blockchain Users, which is now obsolete.

This includes the various classes for the transaction payloads, the user management classes, tests and example programs.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -4201,8 +4197,8 @@ public void completeTx(SendRequest req) throws InsufficientMoneyException {
             checkArgument(!req.completed, "Given SendRequest has already been completed.");
             // Calculate the amount of value we need to import.
             Coin value = Coin.ZERO;
-            boolean requiresInputs = req.tx.getType() != Transaction.Type.TRANSACTION_SUBTX_RESETKEY;
-            boolean requiresNoFee = req.tx.getType() == Transaction.Type.TRANSACTION_SUBTX_RESETKEY;
+            boolean requiresInputs = true;
```
sambarboza: Can't this variable be removed? By reading the method it looks like its value never changes code execution.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -4201,8 +4197,8 @@ public void completeTx(SendRequest req) throws InsufficientMoneyException {
...
+            boolean requiresNoFee = false;
```
sambarboza: Because this variable is always `false` code at lines 4243 and 4244 are never executed. If this code isn't needed, I suggest removing these two lines and the variable itself.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
HashEngineering: It can, but I did find one transaction type that does not require fees or inputs, which is for quorum commitments.  Therefore I will leave this code in for now, but change it to:

```
boolean requiresInputs = req.tx.getType() != Transaction.Type.TRANSACTION_QUORUM_COMMITMENT;
boolean requiresNoFee = req.tx.getType() == Transaction.Type.TRANSACTION_QUORUM_COMMITMENT;
```

core/src/main/java/org/bitcoinj/wallet/Wallet.java
HashEngineering: See above comment

PR #69, opened 2020-02-26
Credit funding Transactions [NMA-385, NMA-386, NMA-387]

HashEngineering:
Adds Credit Funding Transactions to dashj along with support for:
1.  Blockchain Identity Key Chains
2.  Blockchain Identity Funding Key Chains
3.  Fixes to Authentication Key Chains (properly load from wallet file).
4.  Enhancements to Wallet and other KeyChainGroup classes
5.  Adds listeners to Wallet to allow other objects to track the statuses of credit funding transactions, though we may find that other means of tracking are more idea.  That remains to be seen. 

While dashj knows about the key chains, it doesn't know what a Blockchain Identity is. 

This PR should be rebased onto the evonet branch after #67.


core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
```diff
@@ -96,6 +97,7 @@
         names.put(ChainLockSignature.class, "clsig");
         names.put(SendHeadersMessage.class, "sendheaders");
         names.put(GetMasternodePaymentRequestSyncMessage.class, "mnget");
+        names.put(CreditFundingTransaction.class, "tx");
```
HashEngineering: Added to allow a `CreditFundingTransaction` to be sent using `Peer.sendMessage`

The names map says what message ("tx" in this case is associated with what class).

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
```diff
@@ -726,4 +726,11 @@ public int getDIP0008BlockHeight() {
     public int getBIP65Height() {
         return BIP65Height;
     }
+
+    // Coin Type
+    protected int coinType;
+
+    public int getCoinType() {
+        return coinType;
+    }
```
HashEngineering: coinType is 5 for mainnet, 1 for all other testnet and devnets.  This will allow the DerivationPathFactory generate the correct path based on the network.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -842,9 +842,14 @@ public String toString(@Nullable AbstractBlockChain chain, @Nullable CharSequenc
...
-                    s.append(scriptType).append(" addr:").append(scriptPubKey.getToAddress(params));
-                else
+                if (scriptType != null) {
+                    if (scriptType != ScriptType.CREDITBURN)
+                        s.append(scriptType).append(" addr:").append(scriptPubKey.getToAddress(params));
+                    else if (ScriptPattern.isCreditBurn(scriptPubKey)) {
+                        byte [] hash160 = ScriptPattern.extractCreditBurnKeyId(scriptPubKey);
+                        s.append(scriptType).append(" addr:").append(Address.fromPubKeyHash(params, hash160));
+                    }
+                } else
```
HashEngineering: This will label a credit burn output of a transaction when toString() is used.

core/src/main/java/org/bitcoinj/script/Script.java
```diff
@@ -60,7 +60,8 @@
         P2PK(2), // pay to pubkey
         P2SH(3), // pay to script hash
         P2WPKH(4), // pay to witness pubkey hash
-        P2WSH(5); // pay to witness script hash
+        P2WSH(5), // pay to witness script hash
+        CREDITBURN(6); // pay to credit system
```
HashEngineering: Add a credit burn output script type.

core/src/main/java/org/bitcoinj/script/ScriptPattern.java
```diff
@@ -167,6 +167,26 @@ public static boolean isSentToMultisig(Script script) {
...
+        if (chunk0.opcode != OP_RETURN)
+            return false;
+        ScriptChunk chunk1 = chunks.get(1);
+        byte[] chunk1data = chunk1.data;
+        if (chunk1data == null)
+            return false;
+        if (chunk1data.length != 20)
+            return false;
+        return true;
+    }
```
HashEngineering: is the script a credit burn output script:
OP_RETURN pushdata(20) data

core/src/main/java/org/bitcoinj/wallet/AuthenticationKeyChain.java
```diff
@@ -42,6 +46,121 @@
     int currentIndex;
     int issuedKeys;
 
+    public static class Builder<T extends Builder<T>> {
```
HashEngineering: Not sure if we can extend this Builder<> from DeterministicKeyChain.Builder.

core/src/main/java/org/bitcoinj/wallet/DerivationPathFactory.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    // Do we need this?
+    private static DerivationPathFactory instance;
+    public static DerivationPathFactory get(NetworkParameters params) {
+        if(instance == null) {
+            instance = new DerivationPathFactory(params);
+        }
+        return instance;
+    }
```
HashEngineering: This is here to create a singleton of the DerivationPathFactory class.  Not sure if having a singleton is useful, especially in unit tests that may require derivation paths for two different networks.

core/src/main/java/org/bitcoinj/evolution/CreditFundingTransaction.java
```diff
@@ -0,0 +1,212 @@
...
+    public TransactionOutPoint getLockedOutpoint() {
+        if(lockedOutpoint != null)
+            return lockedOutpoint;
+
+        for(int i = 0; i < getOutputs().size(); ++i) {
+            Script script = getOutput(i).getScriptPubKey();
+            if(ScriptPattern.isCreditBurn(script)) {
+                //TODO:  This doesn't look correct, but matches dashsync-iOS.  Why is the txid reversed?
+                //having a reversed txid will not allow it to be searched or matched.
+                lockedOutpoint = new TransactionOutPoint(params, i, Sha256Hash.wrap(getTxId().getReversedBytes()));
```
HashEngineering: This doesn't look right, reversing bytes of transaction hash.  But it is matching the iOS implementation.

core/src/main/java/org/bitcoinj/wallet/AuthenticationKeyChain.java
```diff
@@ -42,6 +46,121 @@
...
+        public T spend(DeterministicKey accountKey) {
+            checkState(accountPath == null, "either spend or accountPath");
+            this.spendingKey = accountKey;
+            this.isFollowing = false;
+            return self();
+        }
+
+        /** The passphrase to use with the generated mnemonic, or null if you would like to use the default empty string. Currently must be the empty string. */
+        public T passphrase(String passphrase) {
+            // FIXME support non-empty passphrase
```
tomasz-ludek: Missed FIXME?

tomasz-ludek:
@HashEngineering,
I noticed that transaction are not being restored after restoring the wallet from seed. Isn't it an issue related to your changes?
For instance I used the wallet created from the following seed:
```
level loan banana buffalo grit donkey weekend true wash hurdle else insane
```
to perform the full flow from ForwardingServiceEvo. At the end the wallet contains 3 transactions:
- incoming funds
- credit funding
- forwarding funds

After deleting the wallet file and restoring the wallet again it contains 0 transactions (after full sync).
Here is how I was restoring wallet:
```
String words = "level loan banana buffalo grit donkey weekend true wash hurdle else insane";
List<String> mnemonicCode = Arrays.asList(words.split(" "));
Date yesterday = new Date(System.currentTimeMillis()-24*60*60*1000);
DeterministicSeed seed = new DeterministicSeed(mnemonicCode, null, "", yesterday.getTime());
kit.restoreWalletFromSeed(seed);
```

tomasz-ludek:
Additionally, it doesn't restore transactions of wallet created on Android (containing multiple standard transactions):
```
hazard holiday airport purchase candy inhale victory athlete melody cement turtle balcony
```
Android wallet (built from `evonet` branch) is able to restore each of mentioned wallets.

HashEngineering:
When I do a restore of a wallet protobuf file, the transactions are found.  This is achieved by deleting the `.spvchain` file which is basically the same thing as a blockchain reset/rescan.

However, using your above code to create the seed and to restore from that seed, the transactions are not found.  My first thought is that the bloom filters are not set up correctly.

HashEngineering:
The android app (on evonet branch) is using a dashj 0.18-SNAPSHOT that does not include the code of this PR.

However, I also built the evonet branch of the app with this PR and these two actions and was able to restore my transactions:
1. blockchain reset/rescan
2. reset wallet / restore from phrase

HashEngineering:
Solution found to the restore issue noted above:

```
String words = "level loan banana buffalo grit donkey weekend true wash hurdle else insane";
List<String> mnemonicCode = Arrays.asList(words.split(" "));
DeterministicSeed seed = new DeterministicSeed(mnemonicCode, null, "", 1427610960l);
kit.restoreWalletFromSeed(seed);
```

HashEngineering:
The creation time of the wallet is used by the peerGroup to manage the syncronization of the blockchain.  Before the creation time, only block headers are retrieved.  After the creation time, the peerGroup will submit the bloom filter and then request merkle blocks from its peers.  Therefore, it would appear without an early enough timestamp for the creation date, then the blockchain sync skips the blocks that in your case had the transactions.

core/src/main/java/org/bitcoinj/wallet/AuthenticationKeyChain.java
HashEngineering: This is a copy/paste from DeterministicKeyChain.  We will need to leave it until upstream (bitcoinj fixes this).  Unfortunately, since I copy pasted this code, now there are three copies of it in our dashj library:  DeterministicKeyChain, FriendKeyChain and AuthenticationKeyChain (here).

core/src/main/java/org/bitcoinj/evolution/CreditFundingTransaction.java
HashEngineering: based on comments from our iOS team the L2 functions of the network (dash platform) may be using the reverse byte order that Satoshi came up with for the bitcoin L1 (P2P protocol).  It drives us nuts every time we have to add code that uses hash values, so much big endian and little endian give us nightmares.

HashEngineering:
This was merged manually here:  https://github.com/dashevo/dashj/commit/21eedfb626162337dc4294ab719e83ad5cce001b

PR #72, opened 2020-05-26
Add ECDH KeyCrypter

HashEngineering:
This will be used by the DashPay library

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/BLSKeyExchangeCrypter.java
+++ b/core/src/main/java/org/bitcoinj/crypto/BLSKeyExchangeCrypter.java
@@ -1,6 +1,5 @@
 /*
- * Copyright 2013 Jim Burton.
- * Copyright 2014 Andreas Schildbach
+ * Copyright 2019 Dash Core Group
  *
  * Licensed under the MIT license (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


core/src/main/proto/wallet.proto
```diff
@@ -434,6 +434,7 @@ message Wallet {
     UNENCRYPTED = 1;                 // All keys in the wallet are unencrypted
     ENCRYPTED_SCRYPT_AES = 2;        // All keys are encrypted with a passphrase based KDF of scrypt and AES encryption
     ENCRYPTED_BLS_KEYEXCHANGE_AES = 3;           // Not used for wallet
+    ENCRYPTED_ECDH_KEYEXCHANGE_AES = 4;          // Not used for wallet
```
HashEngineering, line 437: This change requires that the Protos.java be regenerated by protoc.

HashEngineering: There are many changes in `Protos.java` probably due to an updated `protoc` that actually matches the `protobuf` library used by `dashj`.

core/src/main/java/org/bitcoinj/crypto/BLSKeyExchangeCrypter.java
```diff
@@ -103,85 +76,8 @@ public KeyParameter deriveKey(BLSSecretKey secretKey, BLSPublicKey blsPeerPublic
...
-
-            // Encrypt using AES.
-            BufferedBlockCipher cipher = new PaddedBufferedBlockCipher(new CBCBlockCipher(new AESEngine()));
-            cipher.init(true, keyWithIv);
-            byte[] encryptedBytes = new byte[cipher.getOutputSize(plainBytes.length)];
-            final int length1 = cipher.processBytes(plainBytes, 0, plainBytes.length, encryptedBytes, 0);
-            final int length2 = cipher.doFinal(encryptedBytes, length1);
-
-            return new EncryptedData(iv, Arrays.copyOf(encryptedBytes, length1 + length2));
-        } catch (Exception e) {
```
HashEngineering, line 132: this method (decrypt) and the following encrypt method were refactored into a new class called KeyCrypterAESCBC

core/src/main/proto/wallet.proto
HashEngineering: This probably means that the other protobuf generated files in dashj are using an out of date protobuf compiler

PR #81, opened 2020-07-21
Chain Sync Improvements Phase 1

HashEngineering:
Add topup key chain

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/DerivationPathFactory.java
+++ b/core/src/main/java/org/bitcoinj/wallet/DerivationPathFactory.java
@@ -15,6 +15,10 @@ public class DerivationPathFactory {
         this.coinType = new ChildNumber(params.getCoinType(), true);
     }
 
+    /** blockchain identity registration funding derivation path
+     * m/9'/5'/12' (mainnet)
+     * m/9'/1'/12' (testnet, devnets)
+     */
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/bitcoinj/wallet/DerivationPathFactory.java
```diff
@@ -23,6 +27,22 @@ public DerivationPathFactory(NetworkParameters params) {
...
 
+    /** blockchain identity topup funding derivation path
+     * m/9'/5'/12' (mainnet)
+     * m/9'/1'/12' (testnet, devnets)
+     */
+    public ImmutableList<ChildNumber> blockchainIdentityTopupFundingDerivationPath() {
+        return ImmutableList.<ChildNumber>builder()
+                .add(FEATURE_PURPOSE)
+                .add(coinType)
+                .add(new ChildNumber(13, true))
```
sambarboza, line 38: Why it's 13 here and 12 on the comment?

HashEngineering: Good catch!

HashEngineering: https://github.com/dashevo/dashj/pull/81/commits/03114b9f818b226a8ce0193fcb6fe0bdde5322d8 - fixed here

PR #82, opened 2020-07-28
Chain Sync Improvements: Phase 2

HashEngineering:
Handle the encryption required to pay a contact and to receive payments from a contact.

Add contact keychains to the wallet.

Add wallet methods to get contact related transactions or a contact from a transaction


HashEngineering:
This was merged manually

PR #87, opened 2020-08-21
Merged release-0.17 into evonet branch

tomasz-ludek:


HashEngineering:
#87 was squashed into the `master` branch, so we don't need this PR.

PR #97, opened 2020-11-25
Remove the hardened byte for 256-bit children

HashEngineering:
## What was done?

This is to simplify the key derivation process.  It is a breaking change that goes along with DPP 0.17 changes.

## Breaking Changes

This will change how Contact KeyChains and resulting keys and addresses are derived.  If implemented, all clients must do the same otherwise payments to contacts won't be recieved.

## Checklist:


HashEngineering:
Some tests will fail.

PR #103, opened 2021-01-05
DPP 0.17 Updates

HashEngineering:

Updates necessary to support v17 DPP

## Checklist:


core/src/main/java/org/bitcoinj/core/NetworkParameters.java
```diff
@@ -734,4 +735,8 @@ public int getBIP65Height() {
     public int getCoinType() {
         return coinType;
     }
+
+    public String [] getDefaultMasternodeList() {
```
sambarboza, line 739: We can mark this method abstract and remove its body.

HashEngineering: I can do that, but must also add this "empty" method to all of NetworkParameters derived classes.

There is another way to handle this also, like other getter functions in this class.

sambarboza (approved):

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
sambarboza: You're right, nevermind then.

PR #108, opened 2021-02-27
Updates for v18 DPP

HashEngineering:
## Checklist:


core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
```diff
@@ -415,6 +417,16 @@ public synchronized String toString() {
...
         }
+
+        InstantSendLock islock = getInstantSendlock();
+        if (islock != null) {
+            BLSLazySignature isLockSig = islock.getSignature();
+            if (isLockSig != null) {
+                builder.append(" [sig=" + isLockSig +"]");
+            } else {
+                builder.append(" [sig=null]");
+            }
```
tomasz-ludek, line 428: This "if" doesn't seem to be necessary, when isLockSig is null it should be replaced with "null" string by default.

tomasz-ludek (approved):
Looks good!
Single tiny comment that doesn't have to be addressed.

core/src/main/java/org/bitcoinj/core/TransactionConfidence.java
HashEngineering: That sounds right, but I was getting some `NullPointerException`s earlier so I did it like this.

PR #138, opened 2021-11-02
Core 0.18 Upgrade: Part One

HashEngineering:
## Checklist:


core/src/main/java/org/bitcoinj/core/BitcoinSerializer.java
```diff
@@ -288,7 +288,9 @@ private Message makeMessage(String command, int length, byte[] payloadBytes, byt
         } else if(command.equals("qsendrecsigs")) {
             return new QuorumSendRecoveredSignatures(params);
         } else if(command.equals("islock")) {
-            return new InstantSendLock(params, payloadBytes);
+            return new InstantSendLock(params, payloadBytes, InstantSendLock.ISLOCK_VERSION);
+        } else if(command.equals("isdlock")) {
+            return new InstantSendLock(params, payloadBytes, InstantSendLock.ISDLOCK_VERSION);
```
HashEngineering, line 293: This is the part that will handle received deterministic instant send messages, which are "isdlock".

core/src/main/java/org/bitcoinj/core/MasternodePayments.java
```diff
@@ -21,9 +21,10 @@
     }
 
     int getMinMasternodePaymentsProto() {
-        return context.sporkManager.isSporkActive(SporkManager.SPORK_10_MASTERNODE_PAY_UPDATED_NODES)
-                ? MIN_MASTERNODE_PAYMENT_PROTO_VERSION_2
-                : MIN_MASTERNODE_PAYMENT_PROTO_VERSION_1;
+        //return context.sporkManager.isSporkActive(SporkId.SPORK_10_MASTERNODE_PAY_UPDATED_NODES)
+        //        ?
+        return MIN_MASTERNODE_PAYMENT_PROTO_VERSION_2;
+            //    : MIN_MASTERNODE_PAYMENT_PROTO_VERSION_1;
```
HashEngineering, line 27: need to remove commented code.

core/src/main/java/org/bitcoinj/governance/GovernanceVote.java
```diff
@@ -392,7 +391,7 @@ void relay () {
     public boolean sign(ECKey keyMasternode, PublicKey pubKeyMasternode) {
         StringBuilder strError = new StringBuilder();
 
-        if (context.sporkManager.isSporkActive(SPORK_6_NEW_SIGS)) {
+        /*if (context.sporkManager.isSporkActive(SPORK_6_NEW_SIGS)) {
```
HashEngineering, line 394: remove commented code.

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
```diff
@@ -158,26 +161,30 @@ public String getSporkAddress() {
...
             // A script containing the difficulty bits and the following message:
             //
             //   coin dependent
-            byte[] bytes = Utils.HEX.decode(CoinDefinition.genesisTxInBytes);
+            byte[] bytes = Utils.HEX.decode("04ffff001d01044c5957697265642030392f4a616e2f3230313420546865204772616e64204578706572696d656e7420476f6573204c6976653a204f76657273746f636b2e636f6d204973204e6f7720416363657074696e6720426974636f696e73");
 
             t.addInput(new TransactionInput(n, t, bytes));
             ByteArrayOutputStream scriptPubKeyBytes = new ByteArrayOutputStream();
-            Script.writeBytes(scriptPubKeyBytes, Utils.HEX.decode(CoinDefinition.genesisTxOutBytes));
+            Script.writeBytes(scriptPubKeyBytes, Utils.HEX.decode("040184710fa689ad5023690c80f3a49c8f13f8d45b8c857fbcbc8bc4a8e4d3eb4b10f4d4604fa08dce601aaf0f470216fe1b51850b4acf21b179c45070ac7b03a9"));
```
Syn-McJ, line 184: Are we sure we want to use magic strings here instead of old consts? At least the naming made what's happening a bit more clear. If we don't need `CoinDefinition` anymore, maybe just create some `const` values here or in a new class.

Syn-McJ (approved):
Looks great. I'm a bit concerned about some magic tx strings and numbers, in most places their meaning is clear from variable names, but in some others, it's less so.

core/src/main/java/org/bitcoinj/core/NetworkParameters.java
HashEngineering: Some history: I created `CoinDefinition` about 7 years ago to make it easier to make forks of `bitcoinj` for other coins.  All of the values that make a coin unique would be in one class.  The values within this class are considered as constants since they define parameters that make Dash unique from Bitcoin.  

I thought to remove this class since it was no longer updated and its existence became obsolete.

What I can do is for some of these is make them into `const` values as well as add comments explaining what they are for or where they come from.  This will be done in the next phase of upgrades for Dash Core 0.18.  A note was marked in the story.

The one above `CoinDefinition.genesisTxOutBytes` is only used in one place and is mirrored here in the Dash Core Reference code:  https://github.com/dashpay/dash/blob/master/src/chainparams.cpp#L65-L81

Many of the `CoinDefinition` class fields came from this same C++ source file.

Commit: https://github.com/dashevo/dashj/pull/138/commits/72b3aadfaef43c603d701c0ed840b081ac361115

PR #142, opened 2021-11-16
Allow lower protocol (NMA-1078)

HashEngineering:

`supportsEvolution` is removed because it was no longer needed. This was to support DevNets when they were using an older version of Dash Core than Testnet and Mainnet.  This old version may have been Dash Core 0.15.

`supportsV18` is added to allow the client to interact with Dash Core 0.18 nodes as if this client was a 0.17 client.  By default, this value will be `true`.  But with DashPay, it will be set to `false` for now since Platform 0.21 does not support Core 0.18.

## Checklist:


Syn-McJ (approved):
Looks good

PR #156, opened 2022-02-08
Feature: quorum rotation phase three

HashEngineering:
## Checklist:


core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
```diff
@@ -236,7 +237,7 @@ public void processMasternodeListDiff(@Nullable Peer peer, SimplifiedMasternodeL
         log.info("processing mnlistdiff between : " + quorumState.getMnList().getHeight() + " & " + newHeight + "; " + mnlistdiff);
         lock.lock();
         try {
-            quorumState.applyDiff(peer, headersChain, blockChain, mnlistdiff, isLoadedFromFile());
+            quorumState.applyDiff(peer, headersChain, blockChain, mnlistdiff, isLoadingBootStrap);
```
HashEngineering, line 240: `isLoadedFromFile()` was incorrectly specified

HashEngineering: The idea of the bootstrap is that a file that contains all masternodes and quorums can be loaded from the APK of an app to avoid downloading the entire data.  Instead later, when the client needs to update the masternodes/quorums, it will only need to download the difference compared to the bootstrap.

core/src/main/java/org/bitcoinj/quorums/GetQuorumRotationInfo.java
```diff
@@ -29,7 +29,6 @@
 
 public class GetQuorumRotationInfo extends Message {
 
-    private long baseBlockHashesCount;
     private ArrayList<Sha256Hash> baseBlockHashes;
```
HashEngineering, line 32: The number of `baseBlockHashes` does not need to be stored in a separate variable because the serialization will include the size of that list.

core/src/main/java/org/bitcoinj/quorums/GetQuorumRotationInfo.java
```diff
@@ -50,7 +48,6 @@ public GetQuorumRotationInfo(NetworkParameters params, long baseBlockHashesCount
 
     @Override
     protected void parse() throws ProtocolException {
-        baseBlockHashesCount = readUint32();
         int count = (int)readVarInt();
         baseBlockHashes = new ArrayList<>(count);
         for (int i = 0; i < count; ++i) {
```
HashEngineering, line 53: as seen here, the number of base block hashes included in this message was previously included twice (once as a Uint32 and then as a VarInt.   VarInt, also called compactSize, is the way Bitcoin stores integers using the smallest size possible.  sometimes 1 byte, or 2 bytes, or 3 or 4).

core/src/main/java/org/bitcoinj/store/FlatDB.java
```diff
@@ -194,16 +194,39 @@ ReadResult read(Type object, boolean fDryRun) {
                 magicMessageTmp = new String(vchData, 0, magicMessage.length());
 
                 log.info("file magic message: {}",magicMessageTmp);
+                int fileVersion = 1;
+                try {
+                    String fileVersionString = magicMessageTmp.substring(magicMessageTmp.lastIndexOf('-') + 1);
+                    fileVersion = Integer.parseInt(fileVersionString);
+                } catch (NumberFormatException x) {
+                    //swallow
+                }
```
HashEngineering, line 203: The files dashj uses to store some structures for later have a file version which will allow future version to interpret them correctly.

core/src/main/java/org/bitcoinj/quorums/LLMQParameters.java
```diff
@@ -100,7 +100,7 @@ public static LLMQType fromValue(int value) {
 
         availableLlmqs.put(LLMQType.LLMQ_DEVNET, new LLMQParameters(LLMQType.LLMQ_DEVNET, "llmq_devnet",
                 12, 7, 6, 24, 2, 10,
-                18, 7, 2, 4, 6));
+                18, 7, 4, 4, 6));
```
HashEngineering, line 103: Originally in the `master` branch this value is 3, but previously was 2 and is now 4.  It is not clear if this will change again or return to 3 when this project is done.

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
```diff
@@ -1097,7 +1106,7 @@ public void run() {
                     } catch (IOException x) {
 
                     } catch (BlockStoreException x) {
-
+                        throw new RuntimeException(x);
                     }
                 }
             }
```
Syn-McJ, line 1112: This class is getting big. Not sure if it's the same one that I noted the previous time

Syn-McJ (approved):
Looks good

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
HashEngineering: This is the same file.  Last time you noted this it was 1600 lines, now it is about 1100.

In the next PR, I can refactor further.  Perhaps the bootstrap functionality can go to its own class.

Syn-McJ: Oh ok, great work on refactoring so far.

PR #157, opened 2022-02-15
[NMA-1187] Version 3 Transactions Are Standard

Syn-McJ:
Dash Wallet should accept version 3 transactions immediately.

## Checklist:

- [x] I have performed a self-review of my own code

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -150,7 +150,8 @@ public class Transaction extends ChildMessage {
         }
     }
 
-    public static final int CURRENT_VERSION = 1;
+    public static final int MIN_STANDARD_VERSION = 1;
+    public static final int CURRENT_VERSION = MIN_STANDARD_VERSION;
     public static final int TIMELOCK_VERSION = 2;
     public static final int MAX_STANDARD_VERSION = 3;
 
--- a/core/src/main/java/org/bitcoinj/core/Transaction.java
+++ b/core/src/main/java/org/bitcoinj/core/Transaction.java
@@ -150,6 +150,10 @@ public class Transaction extends ChildMessage {
         }
     }
 
+    public static final int CURRENT_VERSION = 1;
+    public static final int TIMELOCK_VERSION = 2;
+    public static final int MAX_STANDARD_VERSION = 3;
+
     /** Threshold for lockTime: below this value it is interpreted as block number, otherwise as timestamp. **/
     public static final int LOCKTIME_THRESHOLD = 500000000; // Tue Nov  5 00:53:20 1985 UTC
     /** Same but as a BigInteger for CHECKLOCKTIMEVERIFY */
@@ -263,7 +267,7 @@ public class Transaction extends ChildMessage {
 
     public Transaction(NetworkParameters params) {
         super(params);
-        version = 1;
+        version = CURRENT_VERSION;
         inputs = new ArrayList<>();
         outputs = new ArrayList<>();
         // We don't initialize appearsIn deliberately as it's only useful for transactions stored in the wallet.
@@ -272,7 +276,7 @@ public class Transaction extends ChildMessage {
 
     public Transaction(NetworkParameters params, SpecialTxPayload specialTxPayload) {
         this(params);
-        setVersionAndType(3, specialTxPayload.getType());
+        setVersionAndType(MAX_STANDARD_VERSION, specialTxPayload.getType());
         setExtraPayload(specialTxPayload, false);
     }
 
@@ -655,7 +659,7 @@ public class Transaction extends ChildMessage {
         lockTime = readUint32();
         optimalEncodingMessageSize += 4;
 
-        if(getVersionShort() >= 3 && getType() != Type.TRANSACTION_NORMAL) {
+        if(getVersionShort() >= MAX_STANDARD_VERSION && getType() != Type.TRANSACTION_NORMAL) {
             extraPayload = readByteArray();
             setExtraPayloadObject();
             optimalEncodingMessageSize += extraPayload.length;
@@ -756,9 +760,9 @@ public class Transaction extends ChildMessage {
         s.append('\n');
         if (updatedAt != null)
             s.append(indent).append("updated: ").append(Utils.dateTimeFormat(updatedAt)).append('\n');
-        if (version != 1)
+        if (version != CURRENT_VERSION)
             s.append(indent).append("version ").append(version).append('\n');
-        Type type = (getVersionShort() == 3) ? getType() : Type.TRANSACTION_NORMAL;
+        Type type = (getVersionShort() == MAX_STANDARD_VERSION) ? getType() : Type.TRANSACTION_NORMAL;
         s.append("  type ").append(type.toString()).append('(').append(type.getValue()).append(")\n");
         if (isTimeLocked()) {
             s.append(indent).append("time locked until ");
@@ -794,7 +798,7 @@ public class Transaction extends ChildMessage {
         }
         if (!requiresInputs()) {
             // no ins, no outs
-            if (getVersionShort() == 3 && type.isSpecial())
+            if (getVersionShort() == MAX_STANDARD_VERSION && type.isSpecial())
                 s.append(indent).append("payload ").append(getExtraPayloadObject()).append('\n');
 
             return s.toString();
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/bitcoinj/wallet/DefaultRiskAnalysisTest.java
```diff
@@ -253,6 +253,14 @@ public void transactionVersions() throws Exception {
 
         tx.setVersion(3);
         analysis = DefaultRiskAnalysis.FACTORY.create(wallet, tx, NO_DEPS);
+        assertEquals(RiskAnalysis.Result.OK, analysis.analyze());
+
+        tx.setType(Transaction.Type.TRANSACTION_NORMAL);
+        analysis = DefaultRiskAnalysis.FACTORY.create(wallet, tx, NO_DEPS);
+        assertEquals(RiskAnalysis.Result.OK, analysis.analyze());
```
Syn-McJ, line 260: Not sure if we need a separate `version=3; type=normal` test?

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -150,6 +150,11 @@ public boolean isSpecial() {
         }
     }
 
+    public static final int MIN_STANDARD_VERSION = 1;
+    public static final int CURRENT_VERSION = MIN_STANDARD_VERSION;
+    public static final int TIMELOCK_VERSION = 2;
+    public static final int MAX_STANDARD_VERSION = 3;
```
HashEngineering, line 156: I think we will need another constant for version 3 that is used for the rules in Risk Analysis, but rules that pertain to version 3 transactions.

in DIP 2, the version 3 transaction is defined and is called "Special"
https://github.com/dashpay/dips/blob/master/dip-0002.md

So maybe we should have a `DIP2_VERSION = 3`, in a similar way that `TIMELOCK_VERSION = 2`

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -272,7 +277,7 @@ public Transaction(NetworkParameters params) {
 
     public Transaction(NetworkParameters params, SpecialTxPayload specialTxPayload) {
         this(params);
-        setVersionAndType(3, specialTxPayload.getType());
+        setVersionAndType(MAX_STANDARD_VERSION, specialTxPayload.getType());
```
HashEngineering, line 280: This should be `DIP2_VERSION`, since `MAX_STANDARD_VERSION` may be raised to 4 at a later time, but this particular constructor is for creating a version 3 transaction which has a Special transaction payload.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -655,7 +660,7 @@ protected void parse() throws ProtocolException {
         lockTime = readUint32();
         optimalEncodingMessageSize += 4;
 
-        if(getVersionShort() >= 3 && getType() != Type.TRANSACTION_NORMAL) {
+        if(getVersionShort() >= MAX_STANDARD_VERSION && getType() != Type.TRANSACTION_NORMAL) {
```
HashEngineering, line 663: This should be DIP2_VERSION, since MAX_STANDARD_VERSION may be raised to 4 at a later time, but this if block is only execute when reading version 3 transactions.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -756,9 +761,9 @@ public String toString(@Nullable AbstractBlockChain chain, @Nullable CharSequenc
         s.append('\n');
         if (updatedAt != null)
             s.append(indent).append("updated: ").append(Utils.dateTimeFormat(updatedAt)).append('\n');
-        if (version != 1)
+        if (version != CURRENT_VERSION)
             s.append(indent).append("version ").append(version).append('\n');
```
HashEngineering, line 765: The goal here is to append version information if the version is other than 1.  But if CURRENT_VERSION is changed to `2`, then it would add this information for all versions except `2`

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -794,7 +799,7 @@ public String toString(@Nullable AbstractBlockChain chain, @Nullable CharSequenc
         }
         if (!requiresInputs()) {
             // no ins, no outs
-            if (getVersionShort() == 3 && type.isSpecial())
+            if (getVersionShort() == MAX_STANDARD_VERSION && type.isSpecial())
```
HashEngineering, line 802: This should be DIP2_VERSION, since MAX_STANDARD_VERSION may be raised to 4 at a later time.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -876,7 +881,7 @@ public String toString(@Nullable AbstractBlockChain chain, @Nullable CharSequenc
             s.append(indent).append("   fee  ").append(fee.multiply(1000).divide(size).toFriendlyString()).append("/kB, ")
                     .append(fee.toFriendlyString()).append(" for ").append(size).append(" bytes\n");
         }
-        if (getVersionShort() == 3 && type.isSpecial())
+        if (getVersionShort() == MAX_STANDARD_VERSION && type.isSpecial())
```
HashEngineering, line 884: This should be DIP2_VERSION, since MAX_STANDARD_VERSION may be raised to 4 at a later time.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1336,7 +1341,7 @@ static int typeFromLegacyVersion(long version) {
     }
 
     public Type getType() {
-        return versionFromLegacyVersion(version) >= 3 ?
+        return versionFromLegacyVersion(version) >= MAX_STANDARD_VERSION ?
```
HashEngineering, line 1344: This should be DIP2_VERSION, since MAX_STANDARD_VERSION may be raised to 4 at a later time.

core/src/test/java/org/bitcoinj/wallet/DefaultRiskAnalysisTest.java
HashEngineering: normal is the default.  

Though one thing we probably didn't do was put in a Rule in DefaultRiskAnalysis to check the value of type, that it is less than a particular value.  

I suppose we don't need to make a change based on type in this Story.  But, the List of types are here:
https://github.com/dashpay/dips/blob/master/dip-0002-special-transactions.md.  If we did add a check for valid transaction version 3 types, then the type would need to be less than 7.  Though 7 is proposed and won't become official for a few months probably.

HashEngineering (changes requested):
Most changes noted here are for the same reason.

core/src/main/java/org/bitcoinj/core/Transaction.java
HashEngineering: `DIP2_VERSION` could instead be called `SPECIAL_VERSION` -- which seems like a strange name, but `Special` is used in some class names that relate to version 3 transactions.

Syn-McJ: Makes sense. I like "special" more :)

core/src/main/java/org/bitcoinj/core/Transaction.java
Syn-McJ: I see. This was my thinking, but I assumed the goal is to append information if the version isn't the same as the current version.

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -655,7 +662,7 @@ protected void parse() throws ProtocolException {
         lockTime = readUint32();
         optimalEncodingMessageSize += 4;
 
-        if(getVersionShort() >= 3 && getType() != Type.TRANSACTION_NORMAL) {
+        if(getVersionShort() >= SPECIAL_VERSION && getType() != Type.TRANSACTION_NORMAL) {
```
Syn-McJ, line 665: You didn't mention this one, but I think `SPECIAL_VERSION` should be used here as well instead of `MAX_STANDARD_VERSION`?

core/src/main/java/org/bitcoinj/core/Transaction.java
```diff
@@ -1267,7 +1274,7 @@ protected void bitcoinSerializeToStream(OutputStream stream) throws IOException
         for (TransactionOutput out : outputs)
             out.bitcoinSerialize(stream);
         uint32ToByteStreamLE(lockTime, stream);
-        if(getVersionShort() >= 3 && getType() != Type.TRANSACTION_NORMAL) {
+        if(getVersionShort() >= SPECIAL_VERSION && getType() != Type.TRANSACTION_NORMAL) {
```
Syn-McJ, line 1277: You didn't mention this one, but I think `SPECIAL_VERSION` has to be used here as well instead of `MAX_STANDARD_VERSION`?

HashEngineering: Correct, I missed this one.

HashEngineering (approved):
Excellent.

PR #159, opened 2022-03-10
Quorum Rotation: Phase Four

HashEngineering:
## Checklist:


core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -899,14 +896,7 @@ public void startMasternodeListDownload() {
...
             if (context.masternodeListManager.getListAtChainTip().getHeight() < masternodeListBlock.getHeight()) {
-                if (!LLMQUtils.isQuorumRotationEnabled(headerChain, params, params.getLlmqForInstantSend())) {
-                    GetSimplifiedMasternodeListDiff msg = new GetSimplifiedMasternodeListDiff(context.masternodeListManager.getListAtChainTip().getBlockHash(), masternodeListBlock.getHeader().getHash());
-                    sendMessage(msg);
-                } else {
-                    GetQuorumRotationInfo msg = context.masternodeListManager.getQuorumRotationInfoRequest(headerChain.getChainHead());
-                    log.info("message = {}, {}", msg, this);
-                    sendMessage(msg);
-                }
+                context.masternodeListManager.requestQuorumStateUpdate(this, headerChain.getChainHead());
```
HashEngineering, line 899: This is used in DashJ many, many times (`context.masternodeListManager`) and I would like to get away from it.

The context object is effectively a singleton object.  There can only be one context object associated with a thread and only one `NetworkParameters` associated with context.  This limits how many chains an app can be working with (mainnet only, testnet only, etc)

If we wanted an app to support mainnet and testnet at the same time.

HashEngineering: As far as the change made here, this code before the PR is assuming that the DIP4 method of obtaining quorums would be replaced by DIP24.  That is why there is an `if` block.  However, we will need to support DIP4 and DIP24 at the same time.

Additionally, we can say that `Peer` shouldn't care which DIPx is being used to sync masternode lists.  The Masternode List Manager should handle that.

core/src/main/java/org/bitcoinj/evolution/AbstractQuorumState.java
```diff
@@ -0,0 +1,604 @@
...
+    LinkedHashMap<Sha256Hash, StoredBlock> pendingBlocksMap;
+    ArrayList<StoredBlock> pendingBlocks;
+
+    int failedAttempts;
+    static final int MAX_ATTEMPTS = 10;
+
+    public SimplifiedMasternodeListManager.SyncOptions syncOptions;
+    public int syncInterval;
+
+    QuorumStateManager stateManager;
```
HashEngineering, line 95: Most of these class members were taken out of `SimplifiedMasternodeListManager` as part of the refactoring to create a base class that handles most of the similarities between DIP4 and DIP24 quorum list management.

core/src/main/java/org/bitcoinj/evolution/AbstractQuorumState.java
```diff
@@ -0,0 +1,604 @@
...
+
+    public abstract SimplifiedMasternodeList getMasternodeList();
+
+    public abstract SimplifiedMasternodeList getMasternodeListAtTip();
+
+    public abstract LinkedHashMap<Sha256Hash, SimplifiedMasternodeList> getMasternodeListCache();
+
+    public abstract LinkedHashMap<Sha256Hash, SimplifiedQuorumList> getQuorumsCache();
+
+    public abstract ArrayList<Masternode> getAllQuorumMembers(LLMQParameters.LLMQType llmqType, Sha256Hash blockHash);
```
HashEngineering, line 185: Here are the interfaces for what will need to be handled by the subclasses that implement DIP4 and DIP24

core/src/main/java/org/bitcoinj/evolution/AbstractQuorumState.java
```diff
@@ -0,0 +1,604 @@
...
+                                    }
+                                } catch (BlockStoreException x) {
+                                    // do nothing
+                                }
+                            }
+                            if (!pendingBlocksMap.containsKey(block.getHeader().getHash()))
+                                requestMNListDiff(peer, block);
+                        }
+                    }
+                }*/
```
HashEngineering, line 500: further testing is required to find out if this code is needed.

core/src/main/java/org/bitcoinj/evolution/GetSimplifiedMasternodeListDiff.java
```diff
@@ -52,4 +54,31 @@ public boolean equals(Object obj) {
...
+        } catch (BlockStoreException x) {
+            throw new RuntimeException(x);
+        } catch (NullPointerException x) {
+            // swallow
+        }
+
+        return "GetSimplifiedMasternodeListDiff{" +
+                "baseBlockHash=" + baseHeight + " / " + baseBlockHash +
+                ", blockHash=" + blockHeight + " / " + blockHash +
+                '}';
```
HashEngineering, line 82: outputting the block height is very useful when using the logs to debug

core/src/main/java/org/bitcoinj/evolution/QuorumState.java
```diff
@@ -111,12 +140,25 @@ public void applyDiff(Peer peer, AbstractBlockChain headersChain, AbstractBlockC
...
-        mnListsCache.put(newMNList.getBlockHash(), newMNList);
         quorumsCache.put(newQuorumList.getBlockHash(), newQuorumList);
-        mnList = newMNList;
         quorumList = newQuorumList;
     }
 
+    @Override
+    int getBlockHeightOffset() {
+        return SigningManager.SIGN_HEIGHT_OFFSET;
+    }
```
HashEngineering, line 150: With DIP4, our masternode list is always 8 (`SIGN_HEIGHT_OFFSET`) blocks behind the blockchain tip.

core/src/main/java/org/bitcoinj/evolution/QuorumState.java
```diff
@@ -111,12 +140,25 @@ public void applyDiff(Peer peer, AbstractBlockChain headersChain, AbstractBlockC
...
+
+    @Override
+    public int getUpdateInterval() {
+        return 1;
+    }
```
HashEngineering, line 155: With DIP4, every time that a block is added to the blockchain, we will request an update for the masternode/quorum lists.

core/src/main/java/org/bitcoinj/evolution/QuorumState.java
```diff
@@ -161,11 +204,122 @@ public void serializeQuorumsToStream(OutputStream stream) throws IOException {
...
 
+    @Override
+    public SimplifiedMasternodeList getMasternodeList() {
+        return mnList;
+    }
+
+    @Override
+    public SimplifiedMasternodeList getMasternodeListAtTip() {
+        return mnList;
+    }
```
HashEngineering, line 215: We may find we don't need two functions that return the same value.

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
```diff
@@ -476,287 +273,9 @@ public SimplifiedMasternodeList getMasternodeList() {
         }
     }
 
-    public PeerConnectedEventListener peerConnectedEventListener = new PeerConnectedEventListener() {
-        @Override
-        public void onPeerConnected(Peer peer, int peerCount) {
-            lock.lock();
```
HashEngineering, line 482: Most code that was removed from this class was moved to `AbstractQuorumState`

core/src/main/java/org/bitcoinj/evolution/QuorumRotationState.java
```diff
@@ -80,8 +80,8 @@
     SimplifiedQuorumList quorumListAtHMinus3C;
     SimplifiedQuorumList quorumListAtHMinus4C;
 
-    LinkedHashMap<Sha256Hash, SimplifiedMasternodeList> mnListsCache = new LinkedHashMap<>();
-    LinkedHashMap<Sha256Hash, SimplifiedQuorumList> quorumsCache = new LinkedHashMap<>();
+    LinkedHashMap<Sha256Hash, SimplifiedMasternodeList> mnListsCache;
+    LinkedHashMap<Sha256Hash, SimplifiedQuorumList> quorumsCache;
```
HashEngineering, line 84: It seems that in Java if we have initializers like this in the class body and also in constructors, the assignments in the constructors would get replaced by this, causing bugs and crashes due to an empty `LinkedHashMap`.

core/src/main/java/org/bitcoinj/evolution/QuorumRotationState.java
```diff
@@ -864,4 +960,100 @@ public String toString() {
...
+                StoredBlock thisBlock = pendingBlocks.get(0);
+                pendingBlocks.remove(0);
+                pendingBlocksMap.remove(thisBlock.getHeader().getHash());
+            } else log.warn("pendingBlocks is empty");
+
+            if (peer != null && isSyncingHeadersFirst)
+                peer.queueMasternodeListDownloadedListeners(MasternodeListDownloadedListener.Stage.Finished, quorumRotationInfo.getMnListDiffTip());
+
+        } catch (MasternodeListDiffException x) {
+            //we already have this mnlistdiff or doesn't match our current tipBlockHash
```
HashEngineering, line 1000: handling of exceptions is very similar to `QuorumState.professDiff` which can allow for more refactoring

core/src/main/java/org/bitcoinj/evolution/QuorumStateManager.java
```diff
@@ -0,0 +1,12 @@
+package org.bitcoinj.evolution;
```
HashEngineering, line 1: need to add licence

core/src/main/java/org/bitcoinj/evolution/QuorumStateManager.java
```diff
@@ -0,0 +1,12 @@
...
+import java.io.FileNotFoundException;
+
+public interface QuorumStateManager {
+    void save() throws FileNotFoundException;
+    boolean isLoadedFromFile();
+    boolean notUsingBootstrapFile();
+    boolean notUsingBootstrapFileAndStream();
+    void setLoadingBootstrap();
+    void loadBootstrapAndSync();
+}
```
HashEngineering, line 12: This interface allows an `AbstractQuorumState` or derived class to query what is happening with loading and saving files.

`SimplifiedMasternodeListManager` has a single file to store the masternode/quorum state, but there are two systems (`QuorumState` and `QuorumRotationState`) that will need to be saved into that file.

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
```diff
@@ -29,18 +30,14 @@
 import java.util.*;
 import java.util.concurrent.locks.ReentrantLock;
 
-public class SimplifiedMasternodeListManager extends AbstractManager {
+public class SimplifiedMasternodeListManager extends AbstractManager implements QuorumStateManager {
```
Syn-McJ, line 33: So this class is only for the DIP-04 method? I'm wondering if a comment on this can be useful here

Syn-McJ (approved):
Looks great

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
HashEngineering: DIP4 is handled by `QuorumState` and DIP24 by `QuorumRotationState` - This class will make sure that the appropriate DIP's are being handled based on which is active.

HashEngineering: Comments added.

PR #162, opened 2022-04-21
Bugfix: xpub after encryption

HashEngineering:

The xpub of the account key will change after encryption.

## Checklist:

### DIFF
--- a/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
+++ b/core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
@@ -444,9 +444,12 @@ public class DeterministicKeyChain implements EncryptableKeyChain {
         basicKeyChain.importKey(rootKey);
 
         for (int i = 1; i < getAccountPath().size(); i++) {
-            encryptNonLeaf(aesKey, chain, rootKey, getAccountPath().subList(0, i));
+            DeterministicKey parent = hierarchy.get(getAccountPath().subList(0, i - 1), false, false);
+            encryptNonLeaf(aesKey, chain, parent, getAccountPath().subList(0, i));
         }
-        DeterministicKey account = encryptNonLeaf(aesKey, chain, rootKey, getAccountPath());
+        DeterministicKey accountParent = hierarchy.get(getAccountPath().subList(0, getAccountPath().size() - 1), false, false);
+
+        DeterministicKey account = encryptNonLeaf(aesKey, chain, accountParent, getAccountPath());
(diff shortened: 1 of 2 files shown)


Syn-McJ (approved):
Looks good

core/src/main/java/org/bitcoinj/wallet/DeterministicKeyChain.java
```diff
@@ -444,9 +444,12 @@ protected DeterministicKeyChain(KeyCrypter crypter, KeyParameter aesKey, Determi
         basicKeyChain.importKey(rootKey);
 
         for (int i = 1; i < getAccountPath().size(); i++) {
-            encryptNonLeaf(aesKey, chain, rootKey, getAccountPath().subList(0, i));
+            DeterministicKey parent = hierarchy.get(getAccountPath().subList(0, i - 1), false, false);
+            encryptNonLeaf(aesKey, chain, parent, getAccountPath().subList(0, i));
```
HashEngineering, line 448:
```java
encryptNonLeaf(aesKey, chain, rootKey, getAccountPath().subList(0, i));
```
In the original code, the `rootKey` (`M/`) was being used as the parent for all of the children in the account path.

So for `M/44H/5H/0H`, `M` was used as the parent key for 44H, 5H and 0H.  However, the parent of 0H is 5H, not M.  

One could think of this as a tree where each nodes parent was set to root, instead of the actual parent node.

This was an oversight in `bitcoinj`

It turns out that if the account was a BIP32 account, then its path is `M/0H` -- with such a short path, the parents are set properly.  

For any account that has a path size of greater than 2, there will be an issue.

core/src/test/java/org/bitcoinj/wallet/DeterministicKeyChainTest.java
```diff
@@ -625,6 +625,32 @@ public void watchingCannotEncrypt() throws Exception {
...
+        assertEquals(expectedBip32xpub, pub58encrypted);
+
+        // Check the BIP44 keychain
+        DeterministicKey watchingKeyBip44 = bip44chain.getWatchingKey(); //m/44'/1'/0'
+        final String pub58Bip44 = watchingKeyBip44.serializePubB58(params);
+        assertEquals(expectedBip44xpub, pub58Bip44);
+        // Make sure that the xpub remains the same after encrypting the keychain
+        DeterministicKeyChain bip44chainEncrypted = bip44chain.toEncrypted("hello");
+        final String pub58encryptedBip44 = bip44chainEncrypted.getWatchingKey().serializePubB58(params);
+        assertEquals(expectedBip44xpub, pub58encryptedBip44);
```
HashEngineering, line 651: In the current version of dashj, this line will fail, because the xpub would differ because its parent key is set to root instead of the actual parent.

PR #163, opened 2022-04-26
fix: prevent crash if lazy BLS signature is not initialized

HashEngineering:
Signed-off-by: HashEngineering <hashengineeringsolutions@gmail.com>

## Checklist:

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/BLSLazySignature.java
+++ b/core/src/main/java/org/bitcoinj/crypto/BLSLazySignature.java
@@ -5,6 +5,8 @@ import org.bitcoinj.core.NetworkParameters;
 import org.bitcoinj.core.ProtocolException;
 import org.bitcoinj.core.Utils;
 import org.bitcoinj.utils.Threading;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.OutputStream;
@@ -12,6 +14,7 @@ import java.util.concurrent.locks.ReentrantLock;
 
 public class BLSLazySignature extends ChildMessage {
     ReentrantLock lock = Threading.lock("BLSLazySignature");
+    Logger log = LoggerFactory.getLogger(BLSLazySignature.class);
 
     byte [] buffer;
     BLSSignature signature;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/bitcoinj/crypto/BLSLazySignature.java
```diff
@@ -46,7 +49,8 @@ protected void bitcoinSerializeToStream(OutputStream stream) throws IOException
         lock.lock();
         try {
             if (!isSingatureInitialized && buffer == null) {
-                throw new IOException("singature and buffer are not initialized");
+                log.warn("signature and buffer are not initialized");
+                buffer = invalidSignature.getBuffer();
```
HashEngineering, line 53: Not sure why we are throwing an exception here.  Also not sure how the lazy signature is not initialized or has no buffer.

The idea of a lazy signature is to not initialize a BLS Signature object unless it is needed.  This will save some memory usage, though not as much as a BLSLazyPublicKey.

HashEngineering: This crash occurs when obtaining the hash value of the InstantSendLock to print it out to the logs.  We probably need to do a little more investigation to figure out why this crash happened.

```
java.io.IOException: singature and buffer are not initialized
 at org.bitcoinj.crypto.BLSLazySignature.bitcoinSerializeToStream(BLSLazySignature.java:49)
 at org.bitcoinj.core.Message.bitcoinSerialize(Message.java:268)
 at org.bitcoinj.quorums.InstantSendLock.bitcoinSerializeToStream(InstantSendLock.java:99)
 at org.bitcoinj.quorums.InstantSendLock.getHash(InstantSendLock.java:119)
 at org.bitcoinj.quorums.InstantSendManager.processInstantSendLock(InstantSendManager.java:589)
 at org.bitcoinj.quorums.InstantSendManager.processPendingInstantSendLocks(InstantSendManager.java:529)
 at org.bitcoinj.quorums.LLMQBackgroundThread.run(LLMQBackgroundThread.java:36)
```

Syn-McJ (approved):
Looks good

PR #165, opened 2022-07-02
backward compatibilty, bootstraps and the 333 devnet

HashEngineering:
## What was done?

* Refactor to allow for two bootstrap files (mnlistdiff and qrinfo)

## Checklist:


core/src/main/java/org/bitcoinj/core/Context.java
```diff
@@ -377,7 +377,11 @@ public void close() {
             if(masternodeSync.hasSyncFlag(MasternodeSync.SYNC_FLAGS.SYNC_INSTANTSENDLOCKS))
                 llmqBackgroundThread.interrupt();
             blockChain.removeNewBestBlockListener(newBestBlockListener);
-            scheduledMasternodeSync.cancel(true);
+            if (scheduledMasternodeSync != null)
+                scheduledMasternodeSync.cancel(true);
+            blockChain.close();
+            if (headerChain != null)
+                headerChain.close();
```
HashEngineering, line 384: This function is now called by Dash Wallet (another PR) when it disconnects from the network and this will avoid memory leaks where an AbstractBlockChain object remains alive with a closed BlockStore.

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
```diff
@@ -491,99 +515,45 @@ public boolean isLoadedFromFile() {
...
-        Preconditions.checkState(bootStrapFilePath != null || bootStrapStream != null);
-        Preconditions.checkState(getMasternodeList().size() == 0);
-        Preconditions.checkState(getQuorumListAtTip(params.getLlmqChainLocks()).size() == 0);
-        Preconditions.checkState(getMasternodeListCache().size() == 1);
-        Preconditions.checkState(getQuorumListCache(params.getLlmqChainLocks()).size() == 1);
-
-        new Thread(new Runnable() {
-            @Override
-            public void run() {
-
```
HashEngineering, line 542: This was moved into AbstractQuorumState so that each type of State can manage its own bootstrap.  Additionally, the new method doesn't create a thread.

core/src/main/java/org/bitcoinj/core/Context.java
```diff
@@ -291,63 +292,68 @@ public void closeDash() {
...
-        initDashSync(directory, null);
+    public boolean initDashSync(final String directory) {
+        return initDashSync(directory, null);
     }
 
-    public void initDashSync(final String directory, @Nullable String filePrefix) {
-        new Thread(new Runnable() {
-            @Override
-            public void run() {
-                Context.propagate(Context.this);
```
HashEngineering, line 307: here is another place where a thread is removed.  when initDashSync returns, loading of files is complete.

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
```diff
@@ -507,83 +514,39 @@ public void onFirstSaveComplete() {
...
-            }
-        }).start();
+    @Deprecated
+    public static void setBootStrapStream(InputStream mnlistdiffStream, InputStream qrinfoStream, int format) {
+        SimplifiedMasternodeListManager.mnlistdiffBootStrapStream = mnlistdiffStream;
+        SimplifiedMasternodeListManager.qrinfoBootStrapStream = qrinfoStream;
+        bootStrapFileFormat = format;
     }
 
     /**
```
Syn-McJ, line 552: This file is refactored nicely, I think it's this one that was over 1k lines?

core/src/main/java/org/bitcoinj/quorums/InstantSendManager.java
```diff
@@ -404,19 +407,46 @@ boolean processPendingInstantSendLocks() throws BlockStoreException {
...
+
+            for (Sha256Hash islockHash : removed) {
+                pendingInstantSendLocks.remove(islockHash);
+            }
 
             //try to process the invalidInstantSendLocks again
-            for (InstantSendLock isLock : invalidInstantSendLocks.keySet())
-                pendingInstantSendLocks.put(isLock.getHash(), new Pair<>(Long.valueOf(0L), isLock));
+            //for (InstantSendLock isLock : invalidInstantSendLocks.keySet())
+            //    pendingInstantSendLocks.put(isLock.getHash(), new Pair<>(Long.valueOf(0L), isLock));
```
Syn-McJ, line 449: Not needed?

Syn-McJ (approved):
Looks great

core/src/main/java/org/bitcoinj/quorums/InstantSendManager.java
HashEngineering: Good Catch, I will put a TODO on this.  I am not sure.  In the past, this was added to retry ISLocks that had failed, but I am not sure if it ever worked or if it is still needed.

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
HashEngineering: Yes, this was 957 lines before we started Core 18 and it grew considerably by the first time you pointed out the size.

core/src/main/java/org/bitcoinj/quorums/InstantSendManager.java
HashEngineering: One change with Core 0.17 is that when the Blockchain is synchronized, ISLocks are sent for old transactions after we get those transactions.  

This causes a problem, because the masternode lists and quorums that we keep track of are always up to date as of the current time (unless syncing is not complete, then the quorums are out to date).  Old transactions and their islocks will require old masternode lists and quorums to verify, but we don't keep track of those.

What was happening before this was commented out was that we would receive old ISLocks that couldn't be verified, so they would stay in the invalidInstantSendLocks map and be rechecked over and over and over again, and never be verified.

PR #169, opened 2022-08-13
fix: address many masternode/quorum sync issues

HashEngineering:
* handle case of loading inconsistant QuorumState (non-DIP24)
* refactor SimplifiedMasternodeListManager.SyncOptions to MasternodeListSyncOptions
* keep QuorumState synced with blockchain
* Prevent duplicate DIP24 quorum lists from being saved in memory
* check instantsend locks only if qrinfo is synced

## Checklist:

### DIFF
--- a/core/src/main/java/org/bitcoinj/evolution/AbstractQuorumState.java
+++ b/core/src/main/java/org/bitcoinj/evolution/AbstractQuorumState.java
@@ -38,6 +38,7 @@ import org.bitcoinj.core.listeners.PeerDisconnectedEventListener;
 import org.bitcoinj.core.listeners.ReorganizeListener;
 import org.bitcoinj.quorums.LLMQParameters;
 import org.bitcoinj.quorums.QuorumRotationInfo;
+import org.bitcoinj.quorums.SigningManager;
 import org.bitcoinj.quorums.SimplifiedQuorumList;
 import org.bitcoinj.store.BlockStore;
 import org.bitcoinj.store.BlockStoreException;
@@ -46,7 +47,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/bitcoinj/evolution/AbstractQuorumState.java
```diff
@@ -90,7 +90,7 @@
 
     QuorumUpdateRequest<Request> lastRequest;
 
-    static final long WAIT_GETMNLISTDIFF = 5000;
+    static final long WAIT_GETMNLISTDIFF = 5;
```
HashEngineering, line 93: changes made previously are keeping track of the time of sent messages is now using seconds instead of milliseconds.

HashEngineering, line 93: For some reason the units were change to Seconds from Milliseconds

core/src/main/java/org/bitcoinj/evolution/QuorumState.java
```diff
@@ -146,6 +144,10 @@ public void applyDiff(Peer peer, AbstractBlockChain headersChain, AbstractBlockC
             quorumList.syncWithMasternodeList(newMNList);
         }
         if (peer != null && isSyncingHeadersFirst) peer.queueMasternodeListDownloadedListeners(MasternodeListDownloadedListener.Stage.ProcessedQuorums, mnlistdiff);
+
+        // save the current state, if both mnLists and quorums are both applied
+        mnListsCache.put(newMNList.getBlockHash(), newMNList);
+        mnList = newMNList;
         quorumsCache.put(newQuorumList.getBlockHash(), newQuorumList);
         quorumList = newQuorumList;
```
HashEngineering, line 152: found a bug where `mnList` and `quorumList` were at different heights on the blockchain, which might be because the were not "saved" at the same time.  Perhaps an exception with new quorums caused this.  

When these two lists are out of alignment, then it is not possibly to successfully update them without starting over from the beginning of the blockchain or the bootstrap (if there is one).

Syn-McJ: An opportunity to write a test perhaps? :)

Syn-McJ (approved):
Looks good

core/src/main/java/org/bitcoinj/evolution/QuorumState.java
HashEngineering: Ideally, but it won't be easy.

PR #172, opened 2022-08-21
[NMA-1338] CoinJoin P2P messages

Syn-McJ:
## What was done?
- Added classes for the new CoinJoin messages
- Amended `SendDsq` to fit our purposes
- Modified `BitcoinSerialize` to receive the new messages
- Cleaned up some unused/obsolete messages

## Checklist:

- [x] I have performed a self-review of my own code


core/src/main/java/org/bitcoinj/coinjoin/CoinJoinBroadcastTx.java
```diff
@@ -0,0 +1,134 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+
+// dstx
+public class CoinJoinBroadcastTx extends Message {
+    private static final Logger log = LoggerFactory.getLogger(CoinJoinQueue.class);
+
+    private Transaction tx;
+    private TransactionOutPoint masternodeOutpoint;
+    private MasternodeSignature signature;
```
Syn-McJ, line 42: There are different ways to represent the signature, but since this message is received from a Masternode, this type seems suitable, isn't it?

core/src/main/java/org/bitcoinj/coinjoin/PoolMessage.java
```diff
@@ -0,0 +1,57 @@
...
+    ERR_EXISTING_TX(3),
+    ERR_FEES(4),
+    ERR_INVALID_COLLATERAL(5),
+    ERR_INVALID_INPUT(6),
+    ERR_INVALID_SCRIPT(7),
+    ERR_INVALID_TX(8),
+    ERR_MAXIMUM(9),
+    ERR_MN_LIST(10),
+    ERR_MODE(11),
+    ERR_QUEUE_FULL(14),
```
Syn-McJ, line 31: Skipped 12 and 13 since they aren't used anymore.

core/src/main/java/org/bitcoinj/coinjoin/PoolStatusUpdate.java
```diff
@@ -13,10 +13,24 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.bitcoinj.core;
+package org.bitcoinj.coinjoin;
 
-public class DarkSendQueue extends EmptyMessage {
```
Syn-McJ, line 18: `DarkSendQueue` is removed since `dsq` is implemented by `CoinJoinQueue` now.

core/src/main/java/org/bitcoinj/core/DarkSendPool.java
```diff
@@ -1,230 +0,0 @@
...
-
-import java.security.SecureRandom;
-import java.util.ArrayList;
-import java.util.concurrent.locks.ReentrantLock;
-
-/**
- * Created by Eric on 2/8/2015.
- */
-@Deprecated
-public class DarkSendPool {
```
Syn-McJ, line 17: This is not used. Was intended as `CoinJoinStatusUpdate` perhaps? Seems safe to remove.

core/src/main/java/org/bitcoinj/core/Peer.java
```diff
@@ -578,7 +580,7 @@ protected void processMessage(Message m) throws Exception {
             processUTXOMessage((UTXOsMessage) m);
         } else if (m instanceof RejectMessage) {
             log.error("{} {}: Received {}", this, getPeerVersionMessage().subVer, m);
-        } else if(m instanceof DarkSendQueue) {
+        } else if(m instanceof CoinJoinQueue) {
             //do nothing
```
Syn-McJ, line 584: `DarkSendQueue` is replaced with `CoinJoinQueue`, but do we still need to do nothing here?

core/src/main/java/org/bitcoinj/core/SendDsq.java
```diff
@@ -1,12 +1,40 @@
 package org.bitcoinj.core;
 
+import java.io.IOException;
+import java.io.OutputStream;
+
 /**
  * Created by Hash Engineering on 4/12/2019.
  */
-public class SendDsq extends EmptyMessage {
+public class SendDsq extends Message {
```
Syn-McJ, line 9: Should we move this to `/coinjoin` as well?

core/src/main/java/org/bitcoinj/coinjoin/CoinJoinBroadcastTx.java
```diff
@@ -0,0 +1,134 @@
...
+    private long signatureTime;
```
HashEngineering, line 43: this is another place where we will have to have Dash Core (perhaps v19) updated for this message.

`masternodeOutpoint` cannot be linked to a masternode (`SimplifiedMasternodeListEntry`), which is the class that DashJ uses to keep track of masternodes.

core/src/main/java/org/bitcoinj/coinjoin/CoinJoinStatusUpdate.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    protected void parse() throws ProtocolException {
+        sessionID = (int)readUint32();
+        state = PoolState.fromValue((int)readUint32());
+
+        if (protocolVersion <= 702015) {
+            cursor += 4; // Skip deprecated nEntriesCount
+        }
```
HashEngineering, line 58: excellent

core/src/main/java/org/bitcoinj/coinjoin/PoolStatusUpdate.java
HashEngineering: I wonder why git thought this file was moved and changed rather than deleted.

core/src/main/java/org/bitcoinj/core/DarkSendPool.java
HashEngineering: This was added many years ago.  it was used as a background thread to sync the older format of masternode lists, sporks, governance objects (proposals and votes).  Slowly, changes were made to handle these operations in other places.  Masternode Lists synchronization was changed significantly.  Sporks are synced when connected to new peers.  

Last year, this background thread was moved to `Context` using `ScheduledExecutorService` 
https://github.com/dashevo/dashj/commit/ca16fdbbb020bbf82c36c45b6c1a202e1484309c#diff-46be2f806a625e71f6d79476b0f0cf41a615bc1ffdb6911582bc3afc03602549

We only start use `ScheduledExecutorService` if we want to sync governance objects, which we don't do with Dash Wallet.  This eliminates a background thread that wasn't doing anything.

Syncing governance objects would allow Dash Wallet or some other app (we were once working on something called Dash Control) to be set up as masternode operator.  In such a case, the app needs to obtain all the proposals and votes from other masternodes to present a view to the user of the proposals similar to what https://dashcentral.org/budget does.  Then the user could cast votes on proposals.  In our development the furthest we got was obtaining the list of votes and proposals (which took a long time).  At this point, the feature is probably broken.

HashEngineering: In short, we don't need this file anymore.

core/src/main/java/org/bitcoinj/core/Peer.java
HashEngineering: For now we can do nothing.  The goal with this PR is to add the message classes.  We are not sending or receiving or processing any of the messages.

core/src/main/java/org/bitcoinj/core/SendDsq.java
HashEngineering: Let's move it to the `conjoin` package.

HashEngineering (approved):
I think there is only one change to do, which is to move one class to CoinJoin `coinjoin`

core/src/main/java/org/bitcoinj/coinjoin/CoinJoinAccept.java
```diff
@@ -0,0 +1,72 @@
...
+import org.bitcoinj.core.ProtocolException;
+import org.bitcoinj.core.Transaction;
+import org.bitcoinj.core.Utils;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+// dsa
+public class CoinJoinAccept extends Message {
+    private int denomination;
```
HashEngineering, line 29: This may be a set of flags (0x1, 0x2, 0x4 mean different denominations), but we can add constants for it in a subsequent PR.  I don't remember how this is handled in Dash Core.

core/src/main/java/org/bitcoinj/core/SendDsq.java
HashEngineering: Perhaps we should also rename it to `SendCoinJoinQueue` so it will align with `CoinJoinQueue`

core/src/main/java/org/bitcoinj/coinjoin/CoinJoinAccept.java
Syn-McJ: Dash Core seems to have a vector of standard denominations: https://github.com/dashpay/dash/blob/develop/src/coinjoin/coinjoin.h#L366-L372

HashEngineering (approved):
We can merge this to master since this won't affect any Core 18 functionality.

PR #174, opened 2022-09-07
feat: add EC and BLS key support to KeyCrypterAESCBC

HashEngineering:
* add a test that uses a EC private key for AES

## Checklist:

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/KeyCrypterAESCBC.java
+++ b/core/src/main/java/org/bitcoinj/crypto/KeyCrypterAESCBC.java
@@ -119,7 +119,7 @@ public class KeyCrypterAESCBC implements KeyCrypter {
 
     @Override
     public Protos.Wallet.EncryptionType getUnderstoodEncryptionType() {
-        return Protos.Wallet.EncryptionType.ENCRYPTED_BLS_KEYEXCHANGE_AES;
+        return Protos.Wallet.EncryptionType.ENCRYPTED_AES;
     }
 
     @Override
--- a/core/src/main/java/org/bitcoinj/crypto/KeyCrypterAESCBC.java
+++ b/core/src/main/java/org/bitcoinj/crypto/KeyCrypterAESCBC.java
@@ -16,7 +16,10 @@
 
 package org.bitcoinj.crypto;
 
+import com.google.common.base.Preconditions;
+import org.bitcoinj.core.ECKey;
 import org.bitcoinj.core.Utils;
+import org.bitcoinj.wallet.Protos;
 import org.bouncycastle.crypto.BufferedBlockCipher;
 import org.bouncycastle.crypto.InvalidCipherTextException;
 import org.bouncycastle.crypto.engines.AESEngine;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/bitcoinj/crypto/KeyCrypterAESCBC.java
```diff
@@ -30,18 +33,17 @@
...
  * <p>You can use this class to:</p>
  *
- * <p>1) Using a ECDSA private key and a public key, create an AES key that can encrypt and decrypt data.
+ * <p>1) Provide an ECKey or BLSSecretKey as the AES Key.
  * </p>
  *
- * <p>2) Using the AES Key generated above, you then can encrypt and decrypt any bytes using
+ * <p>2) Using the AES Key above, you then can encrypt and decrypt any bytes using
  * the AES symmetric cipher.</p>
  */
```
HashEngineering, line 45: Some of the documentation here applies to derived classes.

core/src/main/java/org/bitcoinj/crypto/KeyCrypterAESCBC.java
```diff
@@ -115,6 +117,31 @@ EncryptedData encrypt(byte[] plainBytes, byte [] iv, KeyParameter aesKey) throws
         }
     }
 
+    @Override
+    public Protos.Wallet.EncryptionType getUnderstoodEncryptionType() {
+        return Protos.Wallet.EncryptionType.ENCRYPTED_BLS_KEYEXCHANGE_AES;
+    }
```
HashEngineering, line 123: We don't have a correct value for this yet.  Need to fix it.

core/src/main/java/org/bitcoinj/crypto/KeyCrypterAESCBC.java
```diff
@@ -115,6 +117,31 @@ EncryptedData encrypt(byte[] plainBytes, byte [] iv, KeyParameter aesKey) throws
...
+
+    @Override
+    public KeyParameter deriveKey(CharSequence password) throws KeyCrypterException {
+        throw new UnsupportedOperationException("use one other deriveKey methods instead");
+    }
```
HashEngineering, line 128: This class and derived classes do not support deriving a key from a password.  Instead this class, keys are "derived" from private keys of different formats.

So the keys are not derived, rather the byte array is wrapped inside a KeyParameter object.

Derived classes can derive a key through other means.

core/src/main/java/org/bitcoinj/wallet/Protos.java
```diff
@@ -16488,6 +16488,10 @@ private Wallet() {
        * <code>ENCRYPTED_ECDH_KEYEXCHANGE_AES = 4;</code>
        */
       ENCRYPTED_ECDH_KEYEXCHANGE_AES(4),
+      /**
+       * <code>ENCRYPTED_AES = 5;</code>
+       */
+      ENCRYPTED_AES(5),
```
HashEngineering, line 16494: Fortunately, with this change, we are only adding an enum value.  No changes are made to the wallet format.  

This enum is meant to specify the type of encryption used in wallet, but the values 3-5 are not used for wallet encryption at this time.

core/src/test/java/org/bitcoinj/crypto/KeyCrypterAESCBCTest.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2013 Jim Burton.
+ * Copyright 2014 Andreas Schildbach
```
Syn-McJ, line 3: Looks like missing Dash here?

Syn-McJ (approved):
Looks good

core/src/test/java/org/bitcoinj/crypto/KeyCrypterAESCBCTest.java
HashEngineering: yes, you are right

PR #176, opened 2022-09-13
Fix many bugs from Core 18 updates

HashEngineering:
## Checklist:


core/src/main/java/org/bitcoinj/evolution/AbstractQuorumState.java
```diff
@@ -201,7 +201,11 @@ public boolean reachedMaxFailedAttempts() {
     public abstract void requestUpdate(Peer peer, StoredBlock block);
 
     public void retryLastUpdate(Peer peer) {
-        peer.sendMessage(lastRequest.getRequestMessage());
+        if (peer != null) {
+            peer.sendMessage(lastRequest.getRequestMessage());
+        } else {
+            log.info("no peer supplied to retry the last update request: {}", lastRequest.request);
+        }
```
HashEngineering, line 208: We were getting an NPE if peer was null.  Not sure of the implications of doing this fix, but other changes should have a null peer if the DashJ has been completely disconnected from the network.  The sync process should resume when reconnecting.

core/src/main/java/org/bitcoinj/evolution/AbstractQuorumState.java
```diff
@@ -702,15 +706,20 @@ public void run() {
...
+                                if (downloadPeer == null) {
+                                    chooseRandomDownloadPeer();
+                                }
                                 retryLastUpdate(downloadPeer);
                             }
                         } finally {
-                            context.peerGroup.getLock().unlock();
+                            if (isLocked) {
+                                context.peerGroup.getLock().unlock();
+                            }
```
HashEngineering, line 722: There is a crash due to tryLock failing above.

core/src/main/java/org/bitcoinj/manager/ManagerFiles.java
```diff
@@ -119,7 +119,7 @@ public void saveNow() throws IOException {
         // but they will serialize (using different temp files).
         if (executor.isShutdown())
             return;
-        log.info("Saving manager file: {}", file.toPath());
+        log.info("Saving manager file: {}", file.getAbsolutePath());
```
HashEngineering, line 122: `toPath()` is not available on Android 6 or 7.  `getAbsolutePath()` is used in many other places in DashJ.

HashEngineering: Perhaps we should have been building DashJ for Java 7 to avoid some of these types of errors.

core/src/main/java/org/bitcoinj/wallet/Protos.java
```diff
@@ -16527,6 +16531,10 @@ private Wallet() {
        */
       public static final int ENCRYPTED_ECDH_KEYEXCHANGE_AES_VALUE = 4;
       /**
+       * <pre>
+       * Not used for wallet
+       * </pre>
+       *
```
HashEngineering, line 16537: Protos.java is a generated file from wallet.proto.  One could ask why we include it in the repo.  In bitcoinj it was removed from the repo (since it was autogenerated), but then those developers decided to add it back.

core/src/main/java/org/bitcoinj/core/AbstractManager.java
```diff
@@ -307,7 +307,9 @@ public void setFilename(String filename) {
     }
 
     public void close() {
-        shutdownAutosaveAndWait();
+        if (vFileManager != null) {
+            shutdownAutosaveAndWait();
+        }
```
HashEngineering, line 312: This may be called more than once in Dash Wallet and then `shutdownAutosaveAndWait` would throw an exception.

Syn-McJ (approved):
Looks good

PR #181, opened 2022-10-14
Feature coinjoin prepare

HashEngineering:

Prepare for coinjoin:
1. add a DIP9 type derivation path for coinjoin keys (used after mixing)
2. Add coinjoin keychain to Wallet and some helper methods
3. Add many classes and methods that to coinjoin

## Checklist:

### DIFF
--- a/core/src/main/java/org/bitcoinj/coinjoin/CoinJoinClientSession.java
+++ b/core/src/main/java/org/bitcoinj/coinjoin/CoinJoinClientSession.java
@@ -83,7 +83,7 @@ public class CoinJoinClientSession extends CoinJoinBaseSession {
         // This still leaves more than enough room for another data of typical CreateDenominated tx.
         List<CompactTallyItem> vecTally = mixingWallet.selectCoinsGroupedByAddresses(true, true, true, 400);
         if (vecTally.isEmpty()) {
-            log.info( "CCoinJoinClientSession::CreateDenominated -- SelectCoinsGroupedByAddresses can't find any inputs!");
+            log.info( "coinjoin: selectCoinsGroupedByAddresses can't find any inputs!");
             return false;
         }
 
@@ -106,7 +106,7 @@ public class CoinJoinClientSession extends CoinJoinBaseSession {
             return true;
         }
 
-        log.info( "CCoinJoinClientSession::CreateDenominated -- failed!");
+        log.info( "coinjoin: createDenominated -- failed!");
         return false;
     }
 
@@ -603,7 +603,7 @@ public class CoinJoinClientSession extends CoinJoinBaseSession {
                 if (context.masternodeListManager.getListAtChainTip().getValidMNsCount() == 0 &&
                         context.getParams().getId() != NetworkParameters.ID_REGTEST) {
                     strAutoDenomResult = "No Masternodes detected.";
-                    log.info("CCoinJoinClientSession::DoAutomaticDenominating -- {}", strAutoDenomResult);
+                    log.info("coinjoin: {}", strAutoDenomResult);
                     return false;
                 }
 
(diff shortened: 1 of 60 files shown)


core/src/main/java/org/bitcoinj/coinjoin/CoinJoinClientSession.java
```diff
@@ -0,0 +1,771 @@
...
+        }
+
+        public T get() {
+            return result;
+        }
+
+        public void set(T result) {
+            this.result = result;
+        }
+    }
```
HashEngineering, line 127: maybe there is a standard class that does this?

core/src/main/java/org/bitcoinj/coinjoin/CoinJoinClientSession.java
```diff
@@ -0,0 +1,771 @@
...
+    public void unlockCoins() {
+        if (!CoinJoinClientOptions.isEnabled())
+            return;
+
+        while (true) {
+            //TRY_LOCK(mixingWallet.cs_wallet, lockWallet);
+            //if (!lockWallet) {
+            //    UninterruptibleSleep(std::chrono::milliseconds{50});
+            //    continue;
+            //}
```
HashEngineering, line 509: not sure how to translate this yet.

core/src/main/java/org/bitcoinj/wallet/Protos.java
```diff
@@ -16415,6 +16415,32 @@ public interface WalletOrBuilder extends
...
+
+    /**
+     * <pre>
+     * coinjoin
+     * </pre>
+     *
+     * <code>repeated .wallet.Key keysCoinJoin = 32;</code>
+     */
+    java.util.List<org.bitcoinj.wallet.Protos.Key> 
+        getKeysCoinJoinList();
```
HashEngineering, line 16427: This is autogenerate by the protobuf compiler.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -607,6 +616,15 @@ public DeterministicKey currentKey(KeyChain.KeyPurpose purpose) {
...
 
+    public DeterministicKey currentCoinJoinKey() {
+        keyChainGroupLock.lock();
+        try {
+            return coinJoinKeyChainGroup.currentKey(KeyChain.KeyPurpose.RECEIVE_FUNDS);
+        } finally {
+            keyChainGroupLock.unlock();
+        }
+    }
+
```
HashEngineering, line 627: This is an example of many functions for coinjoin that are copy/pasted from "regular" functions.  There may be 3 different `currentKey` functions now.  

My idea to remove this duplication is to create a few more classes that will encapsulate other keychains for coinjoin and dashpay related functions.

HashEngineering:
before merging the `feature-coinjoin` branch should be rebased to master.

core/src/main/java/org/bitcoinj/coinjoin/CoinJoinClientSession.java
Syn-McJ: `Thread.sleep()` or something like that?

HashEngineering: Yes, this is true.  From a bigger picture point of view I am wondering which thread would be put to sleep.  if this code is executed in a thread that manages communication with a mixing masternode, then it may cause the peer to disconnect.

This waiting step may actually be one reason why mixing takes a long time:  Waiting for the next block to be mined.

Syn-McJ: I see, makes sense

core/src/main/java/org/bitcoinj/core/TransactionDestination.java
```diff
@@ -0,0 +1,107 @@
...
+import org.bitcoinj.script.Script;
+import org.bitcoinj.script.ScriptPattern;
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Arrays;
+
+/**
+ * Created by Hash Engineering on 8/25/2018.
```
Syn-McJ, line 29: Many classes have this which doesn't seem correct :)

Syn-McJ (approved):
Looks great, although I cannot guarantee I didn't miss issues if there are any :)

core/src/main/java/org/bitcoinj/core/TransactionDestination.java
HashEngineering: Correct, it was a copy paste problem.

PR #184, opened 2022-11-07
tests: add bls batch verifier tests and fix bugs

HashEngineering:
## Checklist:

### DIFF
--- a/core/src/main/java/org/bitcoinj/crypto/BLSBatchVerifier.java
+++ b/core/src/main/java/org/bitcoinj/crypto/BLSBatchVerifier.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2019 Dash Core Group.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.bitcoinj.crypto;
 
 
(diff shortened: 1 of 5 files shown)


HashEngineering:
This PR has some changes that I don't want to use related to not forgetting v2 transactions.  This will be handled by Dash Wallet instead.

core/src/main/java/org/bitcoinj/crypto/BLSBatchVerifier.java
```diff
@@ -174,13 +195,13 @@ private boolean verifyBatchInsecure(HashMap<Sha256Hash, Vector<Pair<MessageId, M
                 }
 
                 if (!aggSig.isValid()) {
-                    aggSig = msg.sig;
+                    aggSig = new BLSSignature(msg.sig);
```
HashEngineering, line 198: We need to make a copy of the signature value rather than its reference.

core/src/main/java/org/bitcoinj/wallet/Wallet.java
```diff
@@ -3394,9 +3397,12 @@ public void cleanup() {
...
                         log.info("Removed transaction {} from pending pool during cleanup.", tx.getTxId());
                     } else {
+                        String reason = String.format("%s%s%s", isAnyOutputSpent ? "spent partially" :"",
+                                isAnyOutputSpent && isISLocked ? " and " : "",
+                                isISLocked ? "instant send locked" : "");
                         log.info(
-                                "Cannot remove transaction {} from pending pool during cleanup, as it's already spent partially.",
-                                tx.getTxId());
+                                "Cannot remove transaction {} from pending pool during cleanup, as it's already {}.",
+                                tx.getTxId(), reason);
```
HashEngineering, line 3405: This needs to be removed from this PR.

Syn-McJ: Is this still relevant?

Syn-McJ (approved):
Looks good, but some of your previous comments might not have been addressed

core/src/main/java/org/bitcoinj/wallet/Wallet.java
HashEngineering: Unfortunately, I forgot why I put this comment.  It is OK to leave it in since it only changes how some information is logged.

HashEngineering: Actually, I put this first comment because it hasn't been tested and there are no unit tests for it.  We should remove this until a another time when it can be tested.

The idea was to prevent v2 transactions from being removed from the wallet.  In Dash Wallet, another solution was used.

PR #195, opened 2023-02-16
Feature core 19 hpmn

HashEngineering:
## Checklist:


core/src/test/java/org/bitcoinj/evolution/ProviderTests.java
```diff
@@ -70,6 +70,7 @@ public class ProviderTests {
 
     @Before
     public void startup() throws UnreadableWalletException {
+        BLSScheme.setLegacyDefault(true);
```
HashEngineering, line 73: with Core 19, there is the introduction of Legacy and Basic Schemes for BLS keys.

Many of the previous tests were written using the Legacy Scheme, so we have to add this to some tests to make sure they pass.

core/src/main/java/org/bitcoinj/core/KeyId.java
```diff
@@ -57,19 +82,16 @@ protected void parse() throws ProtocolException {
...
-
         stream.write(bytes);
     }
 
-    public String toString()
-    {
-        return "KeyId(" + Utils.HEX.encode(bytes) +")";
+    public String toString() {
+        return "KeyId(" + Utils.HEX.encode(Utils.reverseBytes(bytes)) +")";
     }
```
HashEngineering, line 90: As with `Sha256Hash` (a 32 byte hash value), `KeyId` (a 20 byte hash) is also displayed in big endian by Dash Core RPC commands, so we will do the same.

Syn-McJ (approved):
Looks good

PR #200, opened 2023-03-16
feat: support ed25519 masternode keys 

HashEngineering:

This PR adds support for Ed25519 keys.  These keys are based on a different elliptical curve -- the Twisted Edwards 25519 curve.  This twisted curve is not the same kind of curve used by ECKey.  At first, I thought I could reuse the code of ECKey, but replace the Curve parameters with "curve25519".  When I did this, the public keys didn't match the unit tests I used.  Then it became known that curve25519 and Ed25519 are not the same, though apparently, it is possible to convert between the two.

For the `Ed25519Key` class, we will use the `Ed25519*` classes in Bouncy Castle.  The Bouncy Castle classes take 32 byte private keys and 32 byte public keys.  However, according to SLIP-0011, Ed25519 will be using 33 byte public keys, which have a 0x00 byte prepended to the public key.  This will align the public key size with ECKey.

SLIP-0010 also defines how keys are derived from the passphrase or mnemonic seed.  Another difference with Ed25519 is that SLIP-0010 only explains hardened key derivation.  Child keys are always derived from the private key, never from the public key.  This must be related to a property of the secp256k1 elliptic curve that is not present with the Ed25519 twisted curve.  With secp256k1, public keys can be derived from public keys.

## Checklist:


core/src/main/java/org/bitcoinj/crypto/KeyType.java
```diff
@@ -20,7 +20,7 @@ public enum KeyType {
     RSA(0),
     ECDSA(1), // secp256k1
     BLS(2),   // BLS12-381
-    EDDSA(3); // ED25519
+    EdDSA(3); // ED25519
```
HashEngineering, line 23: The `Ed` stands for Edwards (one word) where as EC stands for Elliptic Curve (two words).  

Bouncy Castle also uses `Ed` as the start of all related classes.

core/src/main/java/org/bitcoinj/crypto/bls/BLSDeterministicKey.java
```diff
@@ -94,12 +94,10 @@ public BLSDeterministicKey(ImmutableList<ChildNumber> childNumberPath,
             bytes = serialize(null, false, Script.ScriptType.P2PKH);
             extendedPrivateKey = ExtendedPrivateKey.fromBytes(bytes);
             extendedPublicKey = extendedPrivateKey.getExtendedPublicKey();
-            this.parentFingerprint = (parent != null) ? (int)extendedPrivateKey.getParentFingerprint() : 0;
         } else {
             bytes = serialize(null, true, Script.ScriptType.P2PKH);
             extendedPrivateKey = null;
             extendedPublicKey = ExtendedPublicKey.fromBytes(bytes);
-            this.parentFingerprint = (parent != null) ? (int)extendedPublicKey.getParentFingerprint() : 0;
```
HashEngineering, line 102: This resolved an outstanding issue from the previous PR: https://github.com/dashpay/dashj/pull/199#discussion_r1133025331

These lines are removed as `parentFingerprint` is assigned in line 91.

core/src/main/java/org/bitcoinj/wallet/AnyDeterministicKeyChain.java
```diff
@@ -233,6 +237,7 @@ private int calcDefaultLookaheadThreshold() {
         protected IDeterministicKey spendingKey = null;
         protected ImmutableList<ChildNumber> accountPath = null;
         protected KeyFactory keyFactory = ECKeyFactory.get();
+        protected boolean hardenedKeysOnly = false;
```
HashEngineering, line 240: For some purposes, we only want hardened keys or we can only create hardened keys (Ed25519 only supports hardened key derivation according to SLIP-0010).

core/src/test/java/org/bitcoinj/wallet/authentication/AuthenticationKeyChainGroupTest.java
```diff
@@ -166,8 +188,10 @@ public void serializationTest() throws UnreadableWalletException {
         assertEquals(wallet.currentAuthenticationKey(AuthenticationKeyChain.KeyChainType.MASTERNODE_OWNER),
                 walletCopy.currentAuthenticationKey(AuthenticationKeyChain.KeyChainType.MASTERNODE_OWNER));
 
-        // TODO: for some reason "wallet" has creation times for all BLS keys, so this fails if we compare keys, so lets compare priv key bytes
```
HashEngineering, line 169: This bug was fixed in a previous PR, but the comment remained.

core/src/main/java/org/bitcoinj/wallet/WalletProtobufSerializer.java
```diff
@@ -281,6 +282,13 @@ public Protos.Wallet walletToProto(Wallet wallet) {
             extendedKeyChainBuilder.addAllKey(wallet.getProviderOperatorKeyChain().serializeToProtobuf());
             extendedKeyChains.add(extendedKeyChainBuilder.build());
         }
+        if(wallet.getProviderPlatformOperatorKeyChain() != null) {
+            Protos.ExtendedKeyChain.Builder extendedKeyChainBuilder = Protos.ExtendedKeyChain.newBuilder();
+            extendedKeyChainBuilder.setKeyType(Protos.ExtendedKeyChain.KeyType.EDDSA);
+            extendedKeyChainBuilder.setType(Protos.ExtendedKeyChain.ExtendedKeyChainType.MASTERNODE_PLATFORM_OPERATOR);
+            extendedKeyChainBuilder.addAllKey(wallet.getProviderPlatformOperatorKeyChain().serializeToProtobuf());
+            extendedKeyChains.add(extendedKeyChainBuilder.build());
+        }
```
HashEngineering, line 291: This is a "manual" addition to support the new key type.  

In a separate PR, this serialization will be refactored into the `AuthenticationGroupExtension` class.

core/src/main/java/org/bitcoinj/wallet/AuthenticationKeyChain.java
```diff
@@ -380,10 +413,6 @@ public KeyChainType getType() {
         return type;
     }
 
-    public void setHardenedChildren(boolean hardenedChildren) {
-        this.hardenedChildren = hardenedChildren;
-    }
```
HashEngineering, line 385: this property is being moved to a super class.

Syn-McJ (approved):
Looks great

PR #202, opened 2023-03-30
fix: updates to support Core 19 RC7

HashEngineering:
## Checklist:

### DIFF
--- a/core/src/main/java/org/bitcoinj/evolution/QuorumRotationState.java
+++ b/core/src/main/java/org/bitcoinj/evolution/QuorumRotationState.java
@@ -691,13 +691,18 @@ public class QuorumRotationState extends AbstractQuorumState<GetQuorumRotationIn
         log.info(builder.toString());
     }
 
+    private static ArrayList<ArrayList<SimplifiedMasternodeListEntry>> createNewQuarterQuorumMembers(LLMQParameters llmqParameters) {
+        int nQuorums = llmqParameters.getSigningActiveQuorumCount();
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/bitcoinj/quorums/FinalCommitment.java
```diff
@@ -364,7 +364,7 @@ public boolean verifySizes(LLMQParameters llmqParameters) {
     public Sha256Hash getHash() {
         try {
             UnsafeByteArrayOutputStream bos = new UnsafeByteArrayOutputStream(getMessageSize());
-            bitcoinSerializeToStream(bos, BLSScheme.isLegacyDefault());
+            bitcoinSerializeToStream(bos, isLegacy());
```
HashEngineering, line 367: We should serialize this object for hashing using the BLS Scheme based on the version of the object, rather than the default BLS Scheme which is a global setting.

core/src/main/java/org/bitcoinj/evolution/QuorumRotationState.java
```diff
@@ -804,7 +809,7 @@ private ArrayList<ArrayList<SimplifiedMasternodeListEntry>> buildNewQuorumQuarte
                         // we made full "while" loop
                         if (!updated) {
                             // there are not enough MNs, there is nothing we can do here
-                            return new ArrayList<>(nQuorums);
+                            return createNewQuarterQuorumMembers(llmqParameters);
```
HashEngineering, line 812: This was found to be a misalignment with Core 19.  However, during testnet instability this code was triggered and resulted in failure to verify the membership of one or two quorums.

Syn-McJ (approved):
Looks good

PR #218, opened 2023-06-17
Feature core 19.2 upgrade

HashEngineering:
## Checklist:


core/src/main/java/org/bitcoinj/evolution/ProviderRegisterTx.java
```diff
@@ -81,7 +81,7 @@ protected void parse() throws ProtocolException {
         mode = readUint16();
         collateralOutpoint = new TransactionOutPoint(params, payload, cursor);
         cursor += collateralOutpoint.getMessageSize();
-        address = new MasternodeAddress(params, payload, cursor, NetworkParameters.ProtocolVersion.CURRENT.getBitcoinProtocolVersion());
+        address = new MasternodeAddress(params, payload, cursor, params.getProtocolVersionNum(NetworkParameters.ProtocolVersion.CURRENT));
```
HashEngineering, line 84: I was using the wrong getter for the protocol version in many places.  

When we want to get the current protocol version, we should use:
```
params.getProtocolVersionNum(NetworkParameters.ProtocolVersion.CURRENT)
```
because `getProtocolVersionNum` can be overridden by a network.

core/src/main/java/org/bitcoinj/evolution/QuorumRotationState.java
```diff
@@ -336,8 +333,7 @@ public void applyDiff(Peer peer, AbstractBlockChain headersChain, AbstractBlockC
             SimplifiedQuorumList newQuorumListAtHMinusC = baseQuorumList.applyDiff(quorumRotationInfo.getMnListDiffAtHMinusC(), isLoadingBootStrap, chain, true, false);
 
             baseQuorumList = quorumsCache.get(quorumRotationInfo.getMnListDiffAtH().prevBlockHash);
-            // only verify quorums if extraShare == true, otherwise we don't have all the necessary mn lists
-            SimplifiedQuorumList newQuorumListAtH = baseQuorumList.applyDiff(quorumRotationInfo.getMnListDiffAtH(), isLoadingBootStrap, chain, true, quorumRotationInfo.hasExtraShare());
+            SimplifiedQuorumList newQuorumListAtH = baseQuorumList.applyDiff(quorumRotationInfo.getMnListDiffAtH(), isLoadingBootStrap, chain, true, false);
```
HashEngineering, line 336: This change will skip verifying of old quorums, which adds several seconds of processing on mobile devices.

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeList.java
```diff
@@ -49,16 +50,10 @@ public SimplifiedMasternodeList(NetworkParameters params) {
     SimplifiedMasternodeList(SimplifiedMasternodeList other, short version) {
         super(other.params);
         this.version = version;
-        this.protocolVersion = version == SimplifiedMasternodeListDiff.BASIC_BLS_VERSION ?
-                NetworkParameters.ProtocolVersion.CURRENT.getBitcoinProtocolVersion() :
-                NetworkParameters.ProtocolVersion.BLS_LEGACY.getBitcoinProtocolVersion();
+        this.protocolVersion = params.getProtocolVersionNum(NetworkParameters.ProtocolVersion.SMNLE_VERSIONED);
```
HashEngineering, line 53: when creating a new `SimplifiedMasternodeList`, we will the most recent known protocol version that will be used to save data.

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeList.java
```diff
@@ -49,16 +50,10 @@ public SimplifiedMasternodeList(NetworkParameters params) {
...
         this.blockHash = other.blockHash;
         this.height = other.height;
         mnMap = new HashMap<Sha256Hash, SimplifiedMasternodeListEntry>(other.mnMap);
-        // tell all entries to render as version 2 (BLS basic scheme)
-        for (SimplifiedMasternodeListEntry mn : mnMap.values()) {
-            mn.version = version;
-        }
```
HashEngineering, line 61: changes in Core 19.2 no longer require updating the version of masternode list entries

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeList.java
```diff
@@ -80,7 +75,7 @@ public SimplifiedMasternodeList(NetworkParameters params) {
 
     @Override
     protected void parse() throws ProtocolException {
-        if (protocolVersion >= params.getProtocolVersionNum(NetworkParameters.ProtocolVersion.BLS_SCHEME)) {
+        if (protocolVersion >= NetworkParameters.ProtocolVersion.BLS_SCHEME.getBitcoinProtocolVersion()) {
```
HashEngineering, line 78: For serialization, we need to use particular protocol versions that don't change and cannot be overridden as the serialization must match Dash Core 19.2 exactly.

core/src/main/java/org/bitcoinj/core/MessageSerializer.java
```diff
@@ -39,7 +39,9 @@ public int getProtocolVersion() {
     }
 
     public void setProtocolVersion(int protocolVersion) {
-        this.protocolVersion = protocolVersion;
+        this.protocolVersion = Math.min(protocolVersion, NetworkParameters.ProtocolVersion.CURRENT.getBitcoinProtocolVersion());
+        // the version we select should >= to MINIMUM
+        this.protocolVersion = Math.min(this.protocolVersion, NetworkParameters.ProtocolVersion.MINIMUM.getBitcoinProtocolVersion());
```
Syn-McJ, line 44: Shouldn't the second check be `Math.max` then? With the current check, we always end up with `MINIMUM` as the protocol version, even if `CURRENT` is bigger.

If this is intended, perhaps the comment can be modified to be more clear.

PR #219, opened 2023-06-22
feat: support mainnet upgrade for Core 19.2

HashEngineering:
## Checklist:


core/src/main/java/org/bitcoinj/core/MessageSerializer.java
```diff
@@ -39,9 +39,7 @@ public int getProtocolVersion() {
     }
 
     public void setProtocolVersion(int protocolVersion) {
-        this.protocolVersion = Math.min(protocolVersion, NetworkParameters.ProtocolVersion.CURRENT.getBitcoinProtocolVersion());
-        // the version we select should >= to MINIMUM
-        this.protocolVersion = Math.max(this.protocolVersion, NetworkParameters.ProtocolVersion.MINIMUM.getBitcoinProtocolVersion());
+        this.protocolVersion = protocolVersion;
```
HashEngineering, line 42: In the last PR a bug was found with min that should be max.

However, given that mainnet will be using a lower version of CURRENT protocol at first, this code would set the set the protocol version on messages to 70228 instead of 70227.  

There is probably still a bug here that we can't test for yet.

If we connect to a 70228 node as 70227, then that node will send us 70227 data, but we will interpret it as 70228, which for "mnlistdiff" should give an error.

Therefore, we should connect to nodes as 70228.  70227 nodes will send 70227 data and 70228 nodes send 70228 data.  

But this will probably cause some other problems later.  

It may be better to wait to release Dash Wallet when a significant number of nodes have upgraded to 19.2.

HashEngineering: I also discovered that a single serializer object is being used for all connected nodes, which means that setting the protocol version will cause problems if nodes use different protocol versions.

There were cases where messages were received from two nodes with two protocol versions, both were setting the default single serializer protocol version when the message was received this resulted in a race condition where a 70228 nodes messages were being processed as 70227.

Therefore, for now, we won't set the protocol version until a new serializer is created for each node based on the protocol version.  This will be handled in a new story.

core/src/main/java/org/bitcoinj/core/PeerSocketHandler.java
```diff
@@ -155,8 +155,6 @@ public int receiveBytes(ByteBuffer buff) {
                 Message message;
                 int preSerializePosition = buff.position();
                 try {
-                    // The peer will send us a message based on our protocol version, but theirs may be lower than ours
-                    serializer.setProtocolVersion(peerAddress.protocolVersion);
```
HashEngineering, line 159: This is not thread safe (each connected peer or node is managed by a separate thread, but there is only one serializer object handling all threads and nodes which may have different protocol versions).

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeList.java
```diff
@@ -92,15 +97,10 @@ protected void parse() throws ProtocolException {
...
         size = (int)readVarInt();
-        mnUniquePropertyMap = new HashMap<Sha256Hash, Pair<Sha256Hash, Integer>>(size);
-        for(long i = 0; i < size; ++i)
-        {
-            Sha256Hash hash = readHash();
-            Sha256Hash first = readHash();
-            int second = (int)readUint32();
-            mnUniquePropertyMap.put(hash, new Pair<Sha256Hash, Integer>(first, second));
-        }
+        Preconditions.checkArgument(size == 0, "There is an offset error with this data file, rejecting...");
```
HashEngineering, line 102: This will solve the Out Of Memory exception when loading the mnlist file from storage.  We don't use the mnUniquePropertyMap and save a zero size, but in some cases the mnlist file is corrupted and the size turns out to be non zero and perhaps rather large.  That size is used to create a new HashMap.

Now we will enforce that the size is zero and then fail if it is.  This will result in the mnlist file being ignored and recreated.

core/src/main/java/org/bitcoinj/params/MainNetParams.java
```diff
@@ -348,16 +348,6 @@ public static synchronized MainNetParams get() {
...
 
-    // TODO: Until 19.2 is released on mainnet, we must use 70227 as the min and current version
-    @Override
-    public int getProtocolVersionNum(ProtocolVersion version) {
-        if (version == ProtocolVersion.CURRENT)
-            return ProtocolVersion.CURRENT.getBitcoinProtocolVersion() - 1;
-        else if (version == ProtocolVersion.MINIMUM)
-            return ProtocolVersion.MINIMUM.getBitcoinProtocolVersion() - 1;
-        return super.getProtocolVersionNum(version);
-    }
```
HashEngineering, line 359: Since we will only connect to 19.2 nodes, we don't need to worry about 19.2 nodes that use a lower protocol version.

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeList.java
```diff
@@ -38,9 +40,12 @@ public SimplifiedMasternodeList(NetworkParameters params) {
...
         mnMap = new HashMap<Sha256Hash, SimplifiedMasternodeListEntry>(5000);
-        mnUniquePropertyMap = new HashMap<Sha256Hash, Pair<Sha256Hash, Integer>>(5000);
         storedBlock = new StoredBlock(params.getGenesisBlock(), BigInteger.ZERO, 0);
-        protocolVersion = params.getProtocolVersionNum(NetworkParameters.ProtocolVersion.SMNLE_VERSIONED);
+        initProtocolVersion();
+    }
+
+    private void initProtocolVersion() {
+        protocolVersion = params.getProtocolVersionNum(NetworkParameters.ProtocolVersion.CURRENT);
     }
```
HashEngineering, line 49: We will save in the most up to date protocol version

core/src/main/java/org/bitcoinj/evolution/SimplifiedMasternodeListManager.java
```diff
@@ -314,8 +312,8 @@ public void processMasternodeListDiff(@Nullable Peer peer, SimplifiedMasternodeL
             processQuorumList(quorumState.getQuorumListAtTip());
 
             unCache();
-            // save in the most up-to-date version
-            setFormatVersion(SMLE_VERSION_FORMAT_VERSION);
+            setFormatVersion(QUORUM_ROTATION_FORMAT_VERSION);
```
HashEngineering, line 315: We will save in the latest format version

core/src/test/java/org/bitcoinj/evolution/QuorumStateValidateQuorumsTest.java
```diff
@@ -81,14 +83,16 @@ public static Collection<Object[]> data() {
                         "mnlistdiff-testnet-0-850798-70228-after19.2HF.dat",
                         "testnet-after19.2HF.spvchain",
                         70228,
-                        850798
+                        850798,
+                        SimplifiedMasternodeListManager.SMLE_VERSION_FORMAT_VERSION
```
HashEngineering, line 87: These tests check that we would be saving in the lastest format version.

In the future, I was wondering if the save format should be in protobuf instead of the bitcoin wire protocol format which is easy to mess up with changes from one version to the next.

Syn-McJ (approved):
Looks great

PR #226, opened 2023-08-04
Feature coinjoin improve speed denoms

HashEngineering:
## What was done?

Try these ideas to improve mixing:

    Limit denominations that can be used.  For instance, exclude 0.001.

    Test and use "multi-session"

    Increase the number of active sessions from 4 to 8.

## Checklist:

### DIFF
--- a/core/src/main/java/org/bitcoinj/coinjoin/CoinJoinClientSession.java
+++ b/core/src/main/java/org/bitcoinj/coinjoin/CoinJoinClientSession.java
@@ -552,14 +552,12 @@ public class CoinJoinClientSession extends CoinJoinBaseSession {
         }
 
         try {
+            // don't set the source to TransactionConfidence.Source.SELF on these collateral transactions
+            // otherwise when the masternodes do submit these transactions to the network, they will be
+            // ignored by DashJ and when they are received
             SendRequest req = SendRequest.forTx(txCollateral);
             req.aesKey = context.coinJoinManager.requestKeyParameter(mixingWallet);
             mixingWallet.signTransaction(req);
-            TransactionConfidence confidence = txCollateral.getConfidence();
-            if (confidence != null) {
-                confidence.setSource(TransactionConfidence.Source.SELF);
-                confidence.setConfidenceType(TransactionConfidence.ConfidenceType.UNKNOWN);
-            }
         } catch (ScriptException x) {
             strReason.append("Unable to sign collateral transaction!");
             return false;


core/src/main/java/org/bitcoinj/coinjoin/CoinJoinClientSession.java
```diff
@@ -544,11 +555,19 @@ private boolean createCollateralTransaction(Transaction txCollateral, StringBuil
             SendRequest req = SendRequest.forTx(txCollateral);
             req.aesKey = context.coinJoinManager.requestKeyParameter(mixingWallet);
             mixingWallet.signTransaction(req);
+            TransactionConfidence confidence = txCollateral.getConfidence();
+            if (confidence != null) {
+                confidence.setSource(TransactionConfidence.Source.SELF);
+                confidence.setConfidenceType(TransactionConfidence.ConfidenceType.UNKNOWN);
+            }
```
HashEngineering, line 562: Could this change be related to the fee issue?

.run/WalletTool CJ Mix.run.xml
```diff
@@ -1,8 +1,8 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="WalletTool CJ Mix" type="Application" factoryName="Application">
     <option name="MAIN_CLASS_NAME" value="org.bitcoinj.tools.WalletTool" />
-    <module name="dashj-master-three.tools.main" />
-    <option name="PROGRAM_PARAMETERS" value="mix --wallet=coinjoin.testnet.wallet --net=TEST --debuglog --amount=0.007 --sessions=4 --rounds=1" />
+    <module name="dashj-feature-coinjoin.tools.main" />
+    <option name="PROGRAM_PARAMETERS" value="mix --wallet=coinjoin.testnet.wallet --net=TEST --debuglog --amount=0.15 --sessions=8 --rounds=1 --multi-session" />
```
HashEngineering, line 5: This is the run config that uses "better" parameters to result in faster mixing

core/src/main/java/org/bitcoinj/coinjoin/CoinJoin.java
```diff
@@ -42,12 +43,16 @@
...
+    // CoinJoin::vecStandardDenominations in coinjoin.cpp
+    private static final List<Coin> standardDenominations = Collections.unmodifiableList(
+            Lists.newArrayList(
+                Coin.COIN.multiply(10).add(Coin.valueOf(10000)),
+                Coin.COIN.add(Coin.valueOf(1000)),
+                Coin.COIN.div(10).add(Coin.valueOf(100)),
+                Coin.COIN.div(100).add(Coin.valueOf(10)),
+                Coin.COIN.div(1000).add(Coin.valueOf(1))
+            )
     );
```
HashEngineering, line 56: In a future PR this can be refactored to use `ImmutableList`.

core/src/main/java/org/bitcoinj/coinjoin/CoinJoin.java
```diff
@@ -185,12 +190,14 @@ public static boolean isCollateralValid(Transaction txCollateral, boolean checkI
         }
 
         // the collateral tx must not have been seen on the network
-        boolean hasBeenSeen = txCollateral.getConfidence().getConfidenceType() == TransactionConfidence.ConfidenceType.UNKNOWN;
+        TransactionConfidence confidence = txCollateral.getConfidence();
+        boolean hasBeenSeen = confidence.getConfidenceType() != TransactionConfidence.ConfidenceType.UNKNOWN ||
+                confidence.numBroadcastPeers() > 0 || confidence.getIXType() != TransactionConfidence.IXType.IX_NONE;
```
HashEngineering, line 195: Previously there was a bug here that would flag every collateral tx (used for coinjoin fees) as already spent regardless.  Now we will check 3 things to determine if it was spent yet.

These transactions differ in that they are created by DashJ, but not committed to the wallet and not sent to the network using the normal means of sending transactions.  Rather they are sent to masternodes and will be sent to the network by the mixing masternode when certain conditions are met.  Most collateral tx's are never spent.

core/src/main/java/org/bitcoinj/coinjoin/CoinJoinClientSession.java
```diff
@@ -155,6 +158,13 @@ public int compare(CompactTallyItem o, CompactTallyItem t1) {
         return false;
     }
 
+    public void processTransaction(Transaction tx) {
+        log.info("processing tx of {}: {}", tx.getValue(mixingWallet), tx.getTxId());
+        if (collateralSessionMap.containsKey(tx.getTxId())) {
+            queueTransactionListeners(tx, collateralSessionMap.get(tx.getTxId()), CoinJoinTransactionType.MIXING_FEE);
+        }
+    }
+
```
HashEngineering, line 167: The main point of these `processTransaction` methods is to see if collateral TXes are spent.  If they are, we want to link them back to a Session Id so that it can be determined if the fee was part of the randomly paid fees of successful sessions or if there was a problem with our client and we were charged for bad behavior.  

Currently, DashJ doesn't keep track of this, but it is printed out in a coinjoin-report that can me looked at by the developers.

At this point all fees were charged to successful mixing sessions, which means we don't need to fix bugs related to session failures at this time.

core/src/main/java/org/bitcoinj/coinjoin/CoinJoinClientSession.java
```diff
@@ -203,7 +213,7 @@ private boolean createDenominated(Coin balanceToDenominate, CompactTallyItem tal
         // ****** Add outputs for denoms ************ /
 
         final Result<Boolean> addFinal = new Result<>(true);
-        List<Coin> denoms = CoinJoin.getStandardDenominations();
+        List<Coin> denoms = CoinJoinClientOptions.getDenominations();
```
HashEngineering, line 216: This change will allow the CoinJoin system to use a custom denomination list.

core/src/main/java/org/bitcoinj/coinjoin/CoinJoinClientSession.java
```diff
@@ -110,6 +112,7 @@ public class CoinJoinClientSession extends CoinJoinBaseSession {
 
     private final AtomicBoolean hasNothingToDo = new AtomicBoolean(false); // is mixing finished?
 
+    private final HashMap<Sha256Hash, Integer> collateralSessionMap = new HashMap<>();
```
HashEngineering, line 115: This map will link collateral tx's to the session id, which will make it possible to link a collateral fee tx back to a session.  It could be that the session is `CoinJoinClientSession` has been reset by the time the TX is received.

tools/src/main/java/org/bitcoinj/tools/WalletTool.java
```diff
@@ -1632,10 +1634,11 @@ private static void mix() {
         wallet.getCoinJoin().addKeyChain(wallet.getKeyChainSeed(), DerivationPathFactory.get(wallet.getParams()).coinJoinDerivationPath());
         syncChain();
         // set defaults
-        CoinJoinReporter reporter = new CoinJoinReporter();
+        CoinJoinReporter reporter = new CoinJoinReporter(params);
         CoinJoinClientOptions.setEnabled(true);
         CoinJoinClientOptions.setRounds(4);
         CoinJoinClientOptions.setSessions(1);
+        CoinJoinClientOptions.removeDenomination(CoinJoin.getSmallestDenomination());
```
HashEngineering, line 1641: This is where `WalletTool` will not use the `0.001` denomination.

.run/WalletTool CJ Mix.run.xml
HashEngineering: Using 8 sessions gave performance improvements over 4 sessions.  It could be that 12 or 16 sessions could be faster.

Additionally, using multisession also was improved the speed.

Neither of these values are hard coded anywhere in DashJ.  Currently WalletTool is still using the default settings of Dash Core, which is 4 sessions and multi-session OFF.  However, these will be hard-coded as constants in DashPay in a future story.

core/src/main/java/org/bitcoinj/coinjoin/CoinJoinClientSession.java
HashEngineering: This code resulted in a bug which prevented observing collateral fee tx's.

Syn-McJ (approved):
Great work

PR #266, opened 2024-12-11
Bugfix: sync stuck 33

HashEngineering:

Use longer chain file.

## Checklist:

### DIFF
--- a/core/src/main/java/org/bitcoinj/core/DualBlockChain.java
+++ b/core/src/main/java/org/bitcoinj/core/DualBlockChain.java
@@ -45,9 +45,7 @@ public class DualBlockChain {
 
     public int getBlockHeight(Sha256Hash blockHash) {
         try {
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/bitcoinj/core/DualBlockChain.java
```diff
@@ -72,34 +70,39 @@ public StoredBlock getBlock(Sha256Hash blockHash) {
         }
     }
 
+    private AbstractBlockChain getLongestChain() {
+        return headersChain != null && headersChain.getBestChainHeight() > blockChain.getBestChainHeight() ?
+            headersChain : blockChain;
+    }
```
HashEngineering, line 76: This change, which uses the longer of the two chains, allowed for syncing to get a bit further when there was anuse with a thread freezing on Android:
1. headers - only used to sync headers
2. blocks - used to sync blocks and transactions

This change only delayed when the freeze happened, but this way,  there is a consistant approach.

Syn-McJ (approved):
Looks good

